PR #20, opened 2020-05-26
Apply forceignore rules when checking if folder is empty in Deploy Source task

### LINKED ISSUE #19: Deploy Source task should identify source directory as empty if it only contains ignored files
The Deploy Source task should identify the source directory as empty if it only contains files that are listed in .forceignore

aly76:
#19 

Files matching a pattern in .forceignore are not included in the check for whether a source directory to be deployed is empty.

### DIFF
--- a/packages/core/src/sfdxwrappers/DeploySourceToOrgImpl.ts
+++ b/packages/core/src/sfdxwrappers/DeploySourceToOrgImpl.ts
@@ -184,7 +184,7 @@ export default class DeploySourceToOrgImpl {
       }
     } catch (err) {
       if (err.code === "ENOENT") {
-        throw err; // Re-throw error if .forceignore does not exist
+        throw new Error(`No such file or directory ${err.path}`); // Re-throw error if .forceignore does not exist
       }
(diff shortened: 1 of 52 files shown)


aly76:
Design decisions: 

- Use npm ignore package 
- Relative file paths to the project directory are input to the filter
- Search for .forceignore in the project directory
- Throw error when .forceignore does not exist

packages/core/src/sfdxwrappers/DeploySourceToOrgImpl.ts
```diff
@@ -181,7 +183,10 @@ export default class DeploySourceToOrgImpl {
         return status;
       }
     } catch (err) {
-      if (!this.isToBreakBuildIfEmpty) {
+      if (err.code === "ENOENT") {
+        throw err; // Re-throw error if .forceignore does not exist
```
azlam-abdulsalam: Can we capture and provide a status instead

azlam-abdulsalam (changes requested):
Need version increment for all tasks as its a core change.
Also need to update the task version and changyelog in the documentation

PR #26, opened 2020-05-31
Package Diff implementation

### LINKED ISSUE #21: Package Diff
**Context**
In a mono repo scenario, it would be ideal for the build pipeline  to determine which package has changed and build only those packages rather than rebuilding the entire set of packages consuming resources and increased lead time

**Proposed Solution**
Given a list of source directories, for each directory check whether any changes have occurred since the last successful build for that package, and output the directories  through multiple output variables (for eg: packagename to_be_built or a similar mechanism ) Utilize git based tags as the mechanism to identify whether package has changed.
* The last successful build is identified by the latest tag using a sfpowerscripts defined nomenclature or mention the schema we expect. 
* A change in the package is identified as a file change that does not match a rule in .forceignore

- Pre-Requisite
Every successful build should be tagged (Modify package creation tasks to support it. We would need to enhance the creation tasks)
 

**Similar Solutions**
- Lerna 
- Yarn workspaces

aly76:
Fixes #21 

- Function receives name of package to analyse for changes, and outputs whether the package has changed (true or false). 
- Performs a diff between HEAD and the latest build tag, in the format <package>_v1.0.5.36
- Files matching rules in .forceignore are not included as part of the check

- Added package diff task to CLI


azlam-abdulsalam:
CreateUnlockedPackage has a major version update in the next milestone. So lets stick to it. However CreateSourcePackage Task needs to be updated. Also can we move the git tag function to common, with a split of the git url finder to a seperate function.

PR #27, opened 2020-06-01
Bugfix/ Deploy Source task

aly76:
Deploy Source task is failing in Azure release pipelines, where the project directory is different to the current working directory - thus causing a malformed forceignorePath. 

Fix:  if project directory is defined, then use it to create the forceignorePath.


azlam-abdulsalam (approved):
Looks good

PR #28, opened 2020-06-03
Add support for .next in unlocked packages

### LINKED ISSUE #23: Support for NEXT version in package create task/increment version number
**Context**
Increment Build number task doesn't support the next nomenclature for unlocked packages, thus resulting in couple of issues with traceability when build umber from ci is used, often giving an wrong idea the intermediate package version is lost.

**Proposed Solution**
Add support for next for unlocked/runtime packages. 
- Packages can be determined runtime or unlocked by referring whether a package alias exists in the json file
- Update the buildname to signify a 'next' build is being triggered as the version number is only received after a successful completion of the package create task.

azlam-abdulsalam:
Fixes #23.

@aly76  Need a proper test :)


aly76 (approved):
Okie

PR #48, opened 2020-06-17
Remove tagging function from CLI packaging tasks

aly76:
It is the user's responsibility to push tags that are created as part of packaging tasks in the CLI.


azlam-abdulsalam (approved):
Okie

PR #49, opened 2020-06-19
Utilize Pipeline Decorator Extensibility Point

azlam-abdulsalam:
This update adds  the following
- pre job decorators for Install SFDX, so that that this task is optional
- post job decorator for Create Package Task command, to push tags if all package creation tasks are successful (used in conjunction with diff check)

Please note this PR also bring some changes to the Pipelines


azlam-abdulsalam:
Closing this as pipeline decorators can only be contribued by private extensions

PR #51, opened 2020-06-21
Filter tags against branch and check for changes in package version number

aly76:
- Get tags pointing to the current branch only when checking for changes to package
- Create a new package if the package version number has changed


azlam-abdulsalam:
@aly76  We need to add support for scratch org config file passed in the Create Unlocked Package command.  If the file changes as well, the diff should be enabled.
Also we need to specify the reason why the package is being rebuilt such as
- Change in code
- Change in configuration ( scratch org def / sfdx-project.json)

PR #90, opened 2020-07-14
Save CLI package artifacts to an artifact directory

aly76:
- Add --artifactdir parameter to the Create package commands in the CLI
  - If project directory is specified, the artifact directory is created in the project directory
  - Otherwise the artifact directory will be created in the current working directory
- Create unlocked package no longer manually reads the sfdx-project json to get the package version number
- Default value for --revisionto parameter is HEAD
- Log output variable values
- Fix Delta package path output variable when project directory is specified


azlam-abdulsalam (dismissed):
Okie

PR #109, opened 2020-07-29
Validate code coverage for individual test classes in TriggerApexTest task

### LINKED ISSUE #71: Check individual test class coverage for the Trigger Apex Test task
When the RunSpecifiedTests or RunApexTestSuite parameters are specified, the Trigger Apex Test task should validate that the individual test classes satisfy code coverage requirement.

aly76:


packages/azpipelines/BuildTasks/TriggerApexTestTask/task.json
```diff
@@ -64,6 +64,23 @@
...
+            "helpMarkDown": "When enabled, verifies whether indvidual test classes meet minimum code coverage requirement",
+            "required": false
+        },
+        {
+            "name": "coverageThreshold",
+            "type": "string",
+            "label": "Minimum percentage coverage required per test class",
+            "defaultValue": "75",
+            "helpMarkDown": "Minimum coverage required per test class, in order for the task to succeed",
+            "required":true,
```
azlam-abdulsalam:
```suggestion
            "required":false,
```
Lets not make this a required field

azlam-abdulsalam (changes requested):
Required flag

packages/azpipelines/BuildTasks/TriggerApexTestTask/task.json
aly76: But what if someone leaves out the field?

azlam-abdulsalam: defaultValue will help,  if we make it required, i believe it will create an issue when used in YAML mode as it doesn't respect visiiblity

PR #111, opened 2020-07-29
Fix Trigger Apex Test task dependsOn validatecoverage parameter

aly76:
Remove coveragepercent dependsOn validatecoverage


azlam-abdulsalam (approved):
Okie

PR #129, opened 2020-08-19
Create an artifact folder that contains the source package and artifact metadata

### LINKED ISSUE #126: sfpowerscripts cli generates artifacts on the process working directory
The current cli implementation generates all the artifact metadata  in the process working directory, for unlocked package as well as delta package! This should be modified to support an artifact directory passed on by the user as an input.

aly76:
- Update the CLI packaging commands so that they create an artifact folder that contains the source package and artifact metadata
- Add the artifact directory parameter to the Install Unlocked Package CLI task
- Update the CLI README.md with new parameters 
- Always generate a build artifact in the CLI packaging tasks
- Deprecate artifact_metadata old format in CLI Install Source Package task


azlam-abdulsalam (dismissed):
okie

PR #131, opened 2020-08-23
Refactor Artifact Structure across the package creation tasks

### LINKED ISSUE #56: Add optional flag to deploy a package only if there is a change
Following the recent updates to the plugin, where a package gets created only if there is a change, it would be nice to have an optional flag which prevents a package from being deployed if its already deployed in the last run.

### LINKED ISSUE #101: Remove Confusion around source packages by introducing a new task
Source Packages is an idea to bring artifact based deployment (similar to unlocked packaging) but done off the platform and basically to carry over metadata that is not covered by unlocked packages. It is as simple as having a Json metadata with commit id and a version number, which can then be deployed by metadata api. 
However the use of this tasks are not straight forward, one have to remember, to checkout the artifact and then deploy the directory.

It is better to have a single task that handles installation of source packages

As this will take some time, we need to ensure the current documents are clear on how to deploy source packages

### LINKED ISSUE #126: sfpowerscripts cli generates artifacts on the process working directory

azlam-abdulsalam:
sfpowerscripts feature artifact as the central idea for connecting build and release. Till Milestone 16, the artifacts were only metadata and it need to be checked out from version control, in the case of source package. Delta package did generate the artifacts, but the package format was not fully thought out.
Also to satisfy, skipping of artifacts, if installed not only in unlocked package, also in source package, it is much simpler if we have the artifact carried as part of the build process. This was already done as part of #101, however we have decided to make it uniform across the entire packages.

Every artifact will be named <package_name>_sfpowerscripts_aritfact or spowerscripts_artifact (only in the case of delta package, generated across all the repositories)

Each artifact will contain an aritfact_metadata.json (which has all the information regaring the artifact). as well as a folder called source (which is the metadata that will be deployed). 

Backward compatibility is not supported, The Install Source Package or Deploy Task will only work with the newer versions of associated artifact producer tasks

Fixes #56 
Fixes #101


packages/core/src/sfdxutils/ArtifactGenerator.ts
```diff
@@ -0,0 +1,86 @@
...
+        `${sfdx_package}_sfpowerscripts_artifact`
+      );
+
+      
+      fs.mkdirpSync(sfdx_package_artifact);
+      console.log("Artifact Directory Created at:",sfdx_package_artifact);
+
+      let sourcePackage: string = path.join(
+        sfdx_package_artifact,
+        `source `
```
aly76: There's a trailing blank space after source

PR #136, opened 2020-08-29
Initialise CLI commands by loading sfpowerscripts variables from the environment

### LINKED ISSUE #133: Simplify CLI usage by deprecating readvars
sfpowerscripts cli supports output variables, which are written to the environment of the host. However to consume this, a source readvars has to be issued before any other command. This makes the scripts verbose and defeat the purpose.

sfpowerscripts cli should store this output variable in its own environment config, and all tasks when provided with a non dereferenced  variable, should check the store and figure the value. It should also print the resolved variables

aly76:
Fixes #133 
- CLI commands extend SfpowerscriptsCommand, which contains member function for initialising Sfdx flags with sfpowerscripts environment variables.
- SfpowerscriptsCommand implements `SfdxCommand.run()` as a common entrypoint for all commands, and declares an abstract method `Sfpowerscripts_run()` to be implemented by commands
- Initialise CLI commands by loading sfpowerscripts variables from environment, if the input matches one of the variable names defined in the sfpowerscripts variable dictionary
- Update README.md with recommended way of using output variables
- Decrease indentation in deploy source


azlam-abdulsalam (dismissed):
Okie

azlam-abdulsalam (approved):
okie

PR #142, opened 2020-09-08
Add support for reconcile profiles for source packages

azlam-abdulsalam:
Often profiles has to be reconciled while applying source packages, this becomes a problem while using  the new Install Source Package Task, as the command doesn't offer any support for reconciliation. This means the source code has to be checkout to the particular commit mentioned in the artifact and the tasks added manually. One potential downside to this way, it loses the nice feature of source package in azure pipelines, where source packages which are installed in a org, gets skipped and potentially saving time during release, in a mono repo.

This PR, adds support for reconciliation for source packages, that have "preDeploymentSteps:reconcile" mentioned in sfdx-project.json. 

This is currently added to the InstallSourcePackageTask, however during the subsequent releases, this will be refactored into core  with support for cli


packages/azpipelines/BuildTasks/InstallSourcePackageTask/InstallSourcePackageToOrg.ts
```diff
@@ -141,6 +144,49 @@ async function run() {
     }
 
     //Apply Reconcile if Profiles are found
+    //To Reconcile we have to go for multiple deploys, first we have to reconcile profiles and deploy the metadata
+    let isReconcileActivated = false,isReconcileErrored=false;
+    let profileFolders;
+    if (
+      packageMetadataFromStorage.isProfilesFound &&
+      packageMetadataFromStorage.preDeploymentSteps?.includes("reconcile")
```
aly76, line 152: First time seeing this operator, you learn something new every day !

aly76 (approved):

PR #152, opened 2020-09-21
Amend missing artifact logic to accept array of ArtifactFilePaths

aly76:
- Fixes error with referencing packageMetadataFilePath on an empty array


azlam-abdulsalam (approved):
okie

PR #157, opened 2020-09-29
Fix Trigger Apex Test

aly76:
- Add Apex parsers for identifying test classes & interfaces
- When validating individual class coverage: 
  - Check for classes with no test class
  - Check code coverage of triggers
- Refactor using ManifestHelper


azlam-abdulsalam (changes requested):
Trigger Apex Test task definition should allow for an additional input 'Run only tests in package directory' and this should be the default option 

Also, if this option is selected, upon choosing for validating individual coverage, no additional parameters are required

PR #158, opened 2020-09-30
Refactor Apex parser

aly76:
- Refactor fetchers into a single ApexTypeFetcher that has a function which returns the Apex type (class, test class, interface) of cls files in a directory
- Refactor listeners into single listener that is used to listen for classes, test classes and interfaces


azlam-abdulsalam (approved):
Okie

PR #160, opened 2020-10-01
Refactor/remove apex test suite from source package

azlam-abdulsalam:
- Remove the need for apex test suite for package, Use the apex test class parser to understand the test classes in the package
- Remove the option for destructive manifest in the task (only can be applied through project json)
- Add fallback options to Install Source package to use Local Tests
- Add Option to skip test for Source Packages in Sandboxes
-


packages/azpipelines/BuildTasks/TriggerApexTestTask/task.json
```diff
@@ -9,7 +9,7 @@
     "version": {
         "Major": 9,
         "Minor": 0,
-        "Patch": 0
+        "Patch": 1
```
aly76: Version 9.0.0 TriggerApexTest hasn't been released yet

packages/core/src/sfdxwrappers/CreateSourcePackageImpl.ts
```diff
@@ -114,6 +123,78 @@ export default class CreateSourcePackageImpl {
     return this.packageArtifactMetadata;
   }
 
+  private handleApexTestClasses(mdapiPackage: any) {
+    let apexTypeFetcher: ApexTypeFetcher = new ApexTypeFetcher();
+    let classTypes;
+    try {
+      classTypes = apexTypeFetcher.getApexTypeOfClsFiles(path.join(mdapiPackage.mdapiDir, `classes`));
+    }
+    catch (erorr) {
```
aly76: typo

azlam-abdulsalam: Whats the typo?

azlam-abdulsalam: Ah got it

aly76: Should we also log the error message here. For example if the directory doesn't exist?

packages/core/src/sfdxwrappers/CreateSourcePackageImpl.ts
```diff
@@ -156,9 +237,23 @@ export default class CreateSourcePackageImpl {
...
+  
+    if(fetchedClasses===null || fetchedClasses===undefined)
+        return;
+     
+    let table = new Table({
+      head: ["Class","Path", "Error"],
+    });
 
+    for (let fetchedClass of fetchedClasses) {
+      let item = [fetchedClass.name, fetchedClass.filepath,fetchedClass.error?fetchedClass.error:"N/A"];
```
aly76, line 251: Duplicated fetchedClass.error , or is that the actual syntax ?

aly76: Right it's a ternary operator, got confused without spacing

azlam-abdulsalam: terenary operator,  if null print N/A

packages/core/src/sfdxwrappers/CreateSourcePackageImpl.ts
azlam-abdulsalam: No as source packages without apex are completely okie

PR #169, opened 2020-10-11
Fix SystemRunAs and TestMethod parse errors

aly76:
Fix System.RunAs & TestMethod parse errors by using CaseInsensitiveInputStream instead of ANTLRInputStream


azlam-abdulsalam:
@aly76 Can we remove the Manual checks added?

PR #180, opened 2020-10-19
Check for triggers with no test class

aly76:
If a trigger doesn't get invoked as part of any test class (directly or indirectly), it will not appear in the code coverage json. Therefore, it is necessary to check whether triggers are accounted for in the code coverage json, and mark any missing triggers with zero percent coverage.


azlam-abdulsalam (approved):
okie

PR #182, opened 2020-10-20
Fix TriggerApexTest validate coverage when there are no triggers

aly76:
Fix bug in TriggerApexTest, when validating code coverage, and there are no triggers in the package.


azlam-abdulsalam (approved):
Looks good!

PR #189, opened 2020-10-26
Bug: TriggerApextTestTask not respecting .forceignore when detecting test classes in a package

aly76:


azlam-abdulsalam:
Need to check behaviour in create source Package as well

azlam-abdulsalam (dismissed):
okie

PR #198, opened 2020-11-05
Add Postpack command

### LINKED ISSUE #68: Consolidated Task Executor
In a monorepo, a project will be operating with multiple packages and new packages being added or removed during the course of development.
This means whenever a new package is added, additional tasks that need to be added including in validation (such as PMD, Metadata Validation, Trigger apex tests etc).Same applies for build and deployment stages.

One approach would be is to have a single task, that reads sfdx-project.json and executes the whole set. This has a downside as the logs of such a task is often hard to comprehend, but could be mitigated with folding  ( see grouping in azure pipelines https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#formatting-commands ).
This could be an optional addition to sfpowerkit where a pipeline could be set up and maintained only by changing the project manifest

aly76:
- New CLI command for performing post-packaging tasks: 
  - Unlocked package promotion
  - Zipping
  - Push Git tags
- Removed error thrown within PromoteUnlockedPackageImpl `child_process` callback function
  - Unable to catch error thrown in callback function
- Throw error in ArtifaceFilepathFetcher when the artifact directory does not exist

### DIFF
--- a/packages/core/src/artifacts/ArtifactFilePathFetcher.ts
+++ b/packages/core/src/artifacts/ArtifactFilePathFetcher.ts
@@ -14,6 +14,11 @@ export default class ArtifactFilePathFetcher {
     artifactDirectory: string,
     sfdx_package?: string
   ): ArtifactFilePaths[] {
+
+    if (!fs.existsSync(artifactDirectory)) {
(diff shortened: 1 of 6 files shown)


azlam-abdulsalam (approved):
Okie

PR #199, opened 2020-11-06
Fix child process callback

aly76:
* Errors thrown in the child process callback do not cause the process to exit with status > 0

* In the past, OnExit function rejected promise on 'exit' signal from the child_process, so the callback did not get a chance to run. But since OnExit now listens to 'close' signal, the callback gets a chance to run.
 
* Instead of throwing error in the callback, log stderr to the console. OnExit function will reject promise on non-zero exit code and cause the process to exit.


azlam-abdulsalam (approved):
okie

PR #201, opened 2020-11-09
Move child process stderr listener outside of callback

aly76:
The stderr listener is not getting a chance to run from the child process callback, or there's no data event at that point. Results in the error message not being logged to the console.


azlam-abdulsalam (approved):
Okie

PR #203, opened 2020-11-09
Fix package diff for data packages

aly76:
* Do not apply forceignore when creating data packages
  * CSV files will typically be in the forceignore because it breaks source conversion


azlam-abdulsalam (approved):
okie

PR #214, opened 2020-11-13
Fix bug in postpack command

aly76:
* Assign artifacts length to a variable to prevent changing loop size
* Exit with overall status equal to 1 when a package fails to promote


azlam-abdulsalam (approved):
okie

PR #237, opened 2020-11-25
Refactor artifactFilePathFetcher

aly76:
- Refactor artifactFilePathFetcher()
  - Only unzips single package if sfdx_package is specified
  - Globs for sfpowerscripts_artifact
  - Verify that the artifact filepaths exist on the file system
  - Use latest artifact if more than one is found for the same package
  - Unzip artifacts to .sfpowerscripts directory
 
- Changes to ArtifactGenerator 
  - Substitute Build Number with Pre-release in artifact filename


azlam-abdulsalam (approved):
okie

PR #258, opened 2020-12-10
Update alias value in post manage scratch org

### LINKED ISSUE #235: Created scratch org with a custom alias does not get deleted at the end of the pipeline
When using the "Create / Delete a Scratch Org" task and specifying an alias that is not the default of "scratchorg" and setting maintainorg to "delete" the scratch org is not deleted in the automatically created "Post-job: Create scratch org" task.

It reports the following error:
```
ERROR running force:org:delete:  No org configuration found for name scratchorg
##[error]Command failed: npx sfdx force:org:delete -u  scratchorg -v MCDDevHubOrg -p
```
It appears to ignore the name of the scratch org created and looks for the default of **scratchorg**.

Caitlyn-Mills:
Fixes a bug where the correct scratch org alias was not passed to 'Post Manage Scratch Org task' for scratch org deletion when a custom name is given. 
Closes #235


azlam-abdulsalam (approved):
Okie

PR #259, opened 2020-12-11
Add missing flag option to skip optimized deployments for Source Packages

azlam-abdulsalam:
Add isSkipOptimzedDeployment option to Source Packages
- By default, all source packages will be deployed with optimized deployment, 
which means each package need to have its individual apex classes with 75% or more 
code coverage, Sometime this is not feasible and one cause the whole org's coverage
to push a package successfully to production

This flag was not respected by DeployImpl. This PR fixes it


azlam-abdulsalam (commented):
oke

PR #260, opened 2020-12-11
Get latest commits instead of tags in ValidateImpl

aly76:
- Get latest commits from sfpowerscripts artifact, rather than the tags, to determine which packages need to be validated.
- Assign packageMetadata.tags


azlam-abdulsalam (approved):
okie

PR #275, opened 2020-12-14
Map forceignore file to stage using project config 

aly76:
- Map the forceignore file to be used in stages (prepare, validate, quickbuild, build) in the project config, under plugins -> sfpowerscripts -> ignoreFiles

-  The root forceignore in the artifact is replaced by the appropriate forceignore file at runtime for the prepare & validate commands.

- The root forceignore is replaced before creating a new unlocked package version for quickbuild and build stages.


azlam-abdulsalam (dismissed):
okie

PR #280, opened 2020-12-16
Add source code for sfpowerscripts-artifact package 

aly76:
- Add source code for sfpowerscripts-artifact package
- Pass environment variable to set sfpowerscripts-artifact version in prepare command.


azlam-abdulsalam (approved):
okie

PR #284, opened 2020-12-17
Refactor pre & post-deployment steps

aly76:
- Removed 'preDeploymentSteps' & 'postDeploymentSteps' from packageMetadata 
- Replaced 'permissionSetsToAssign' with 'assignPermSetsPreDeployment' & 'assignPermSetsPostDeployment'
  - String arrays
- reconcileProfiles ON by default for InstallSourcePackage
  - Users can turn it off by explicitly defining reconcileProfiles: false in the project config
- Specify 'preDeploymentScript' and 'postDeploymentScript,' in the project config, to run before/after installation  
- Convert 'skipDeployOnOrgs' from string to string array
- Support for pre & post-deployment permission set assignment across source, data and unlocked packages

### DIFF
--- a/packages/core/src/PackageMetadata.ts
+++ b/packages/core/src/PackageMetadata.ts
@@ -27,8 +27,6 @@ export default interface PackageMetadata {
     metadataCount?:number;
     sourceDir?:string;
     dependencies?:any;
-    preDeploymentSteps?:string[];
-    postDeploymentSteps?:string[];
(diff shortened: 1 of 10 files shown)


azlam-abdulsalam (commented):
We need to move permsets assignment to Install Package Base in the case of CLI, as its common for all classes.

Also azp is missing the features,as its installation we would need it on azp layer as well

PR #302, opened 2021-01-11
Turn off rename detection in PackageDiffImpl

aly76:
Switch off rename detection, so that renames are represented by a deletion and addition of a file. This way, a change to the source package will also be picked up.


azlam-abdulsalam (approved):
Looks good, Lets get a patch out for this one!

PR #308, opened 2021-01-13
Create new orchestrator:validateAgainstOrg command

aly76:
- New command `orchestrator:validateAgainstOrg` allows validation against an user-provided scratch org 
- Added alias `orchestrator:validateAgainstPool`


azlam-abdulsalam (dismissed):
okie

PR #314, opened 2021-01-15
Check package descriptor for changes, in PackageDiffImpl

### LINKED ISSUE #303: orchestrator:build not triggering  a package to be built when there is a change only in sfpowerscripts properties
sfpowerscripts add addtional properties to the packages such as for deploying permsets. Changes only to this particular attribute without any changes to underlying package doesn't get detected as a change

- Workarounds
   - Change Version Number of the package (major/minor/patch)
   -  Introduce a dummy change in any of the included config files

aly76:
- Check package descriptor changes, in PackageDiffImpl
  - Replaces old logic that checked for change in package version only 
- Add unit tests for PackafeDiffImpl
  - Created Util classes for simple git and ignore, to enable mocking


azlam-abdulsalam (dismissed):
Okie

azlam-abdulsalam:
I think we meed more test cases here, considering its a core component, some negative scenarios

azlam-abdulsalam (dismissed):
Okie

PR #331, opened 2021-01-22
Support for artifacts with different source version in DeployImpl

### LINKED ISSUE #332: Orchestrator commands to support non linear pipeline.
CD pipelines are always recommend to be separated from CI pipelines, that means packages generated in a CI pipeline should be aggregated and then deployed to an org without driven by the source code (as Project config determines the order of deployment) . So deploy command when provides with a set of artifacts should determine the order automatically

aly76:
Previously, DeployImpl used the package manifest in the root directory to determine the deployment order. Now, the package manifest is taken from the latest artifact. 

- Artifacts no longer have to be created from the same source _version_
- Artifacts must be created from the same source _repository_ 
- Artifacts carry the original sfdx-project.json at manifests/sfdx-project.json.ori

Refactor:
- Fetch / Unzip artifacts once only

### DIFF
--- a/packages/core/src/artifacts/ArtifactFilePathFetcher.ts
+++ b/packages/core/src/artifacts/ArtifactFilePathFetcher.ts
@@ -8,7 +8,7 @@ import semver = require("semver");
 export default class ArtifactFilePathFetcher {
   /**
    * Decider for which artifact retrieval method to use
-   *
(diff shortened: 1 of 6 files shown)


azlam-abdulsalam (approved):
Oke

PR #337, opened 2021-01-24
Fix bug in PackageDiffImpl when package path contains . cwd segment

aly76:
Context:
Some repositories specify package paths with the cwd '.' segment
e.g. './path/to/package'

Solution: 
Normalize package path before checking whether modified files includes the path
e.g. './path/to/package' -> 'path/to/package'


azlam-abdulsalam:
This needs to be released as a hotfix

PR #344, opened 2021-01-27
Tag only when build succeeds

aly76:
- For consistency, only tag packages if the build succeeds and there are no errors
- Throw error if there is an error creating artifacts


azlam-abdulsalam (approved):
Okie

PR #346, opened 2021-01-28
Fall back to project config in project directory if artifacts do not contain sfdx-project.json.ori

aly76:
- If none of the artifacts contain a sfdx-project.json.ori, then fall back to the project config in the project directory.

- If cannot find project config in the project directory, then throw error.


azlam-abdulsalam (approved):
I am good with this,

PR #351, opened 2021-01-28
Fix closing log group position in DeployImpl

aly76:
- Fix closing log group position in DeployImpl
- Add log group for tests


azlam-abdulsalam (approved):
okie

PR #357, opened 2021-02-02
Remove generate password in ScratchOrgUtils

aly76:
- Remove generate password  and shareScratchOrgThroughEmail in ScratchOrgUtils
- Use auth plugin


azlam-abdulsalam (dismissed):
Okie

PR #360, opened 2021-02-04
Package Build Collections

### LINKED ISSUE #361: Build Packages if any packages in a provided collection changes
For referential integrity of certain packages ( especially data packages') , the packages has to be built if their sibling (the packages in the collection) has any change.  In this particular scenario,  Given  Packages A  and B, if  Package A changes, Package B has to be built, so that these two packages can then be further deployed into an org.

`buildCollection: [ <pkg1> ,  <pgk2> ] ` will be an additional descriptor that is added to the sfdx-project.json .

aly76:
Packages within the same collection should all be built if at least one of the packages in the collection has changed.

New Classes:
- BuildCollections
- UndirectedGraph


azlam-abdulsalam (dismissed):
Nice! Lets get this rolled out

PR #399, opened 2021-02-24
Deprecate GitTag in Build/QuickBuild

### LINKED ISSUE #392: Add support for pushing tags upon completion of build
At present, build commands support creating git tags, but it doesn't support pushing these tags back to the repo.  sfpowerscripts should provide an option to push tags to the repo upon successful completion of the build

### LINKED ISSUE #396: Need mechanism to create tags for publish/promote
Currently build/quickbuild creates tags while building packages, and if the user is pushing tags to the repo at this point at the end of the stage, and say promotion/publish fails (if run on a separate stage on a different instance) , future builds will never pick up any other packages successfully promoted but never published

The solution is to not to push tag such builds and delay the tags till publish, however, publish and promote doesn't provide a mechanism to create tags

azlam-abdulsalam:
GitTag option in Build and QuickBuild often creates confusion on where
to tag and also causes an issue when packages fail to promote such as
below issues

To counter this, it is better to deprecate tagging support in Build
and QuickBuild and move to publish which is the logical endpoint
for journey of a package

- Fixes #396
- Fixes #392


aly76 (approved):
LGTM

PR #400, opened 2021-02-25
Fix Logs to display errors if artifacts are not found in the org

azlam-abdulsalam:
Fix Error Logs being displayed when sfpowerscripts_artifact is not installed﻿


aly76 (approved):
LGTM

PR #409, opened 2021-03-03
Impact analysis on Validate

### LINKED ISSUE #419: Impact Analysis during pull request
Generate impact dependency map of all the components that are getting changed in a pull request.

aly76:
This is a proposed enhancement to validate command to do an impact analysis assessment on every Pull Request. At this stage, it can compare against a given branch, utilizes the proposed changes in a give PR, creates an impact analysis artifact, which is a collection of images/graphs and lives as a build artifact

 This utilizes a variety of libraries to get it done
- sfdc-soup for dependency analysis
- handlebars to create HTML templates
- cytoscape.js for rendering graphs
- puppeteer to create images

There is further work being carried out, as in introspecting whether some of the functionality would make into the main line and might be replaced by a proposed Salesforce app, which sfpowerscripts could use it to publish reports, rather than as build artifacts.


azlam-abdulsalam (approved):
merging it temporarily till we establish the value proposition. 

May not make into Milestone 21

PR #411, opened 2021-03-04
Show text exec error instead of missing test report error

aly76:
Fix Error Message if a test report is missing


azlam-abdulsalam (approved):
okie

PR #413, opened 2021-03-04
Add Log Based Metrics

### LINKED ISSUE #362: Log Based Metrics
sfpowerscripts currently reports build and deployment metrics using statsD. While this is an excellent solution, it been found with multiple customers having a statsD server integration with a cloud based CI/CD (where agents run on the cloud) is very hard. 

Inorder to report this, one  need to utilise log based metrics that basically outputs the current statsD metrics into the ~~console~~ file logs. These build logs could then be transported to a blob storage, from where it could be fetched for analysis like Splunk.

azlam-abdulsalam:
Fixes #362
- Metrics will be logged as JSON
- Logs can be collected from CI/CD systems and then parsed for various metrics﻿


aly76 (approved):
LGTM

PR #423, opened 2021-03-10
Publish artifacts to a NPM registry

### LINKED ISSUE #418: Add an option to create npm compatible packages.

aly76:
#418


azlam-abdulsalam:
Can we get tests updated?

azlam-abdulsalam:
Also , will this work with prepare command's ability to pull artifacts?

PR #438, opened 2021-03-15
All package types must be promoted to be deployed to production

### LINKED ISSUE #428: Fail early during deployment to production if using non promoted packages
Salesforce doesn't allow non promoted unlocked packages to be deployed to prod. If an non promoted package is added to the deployment list, it fails during deployment leaving the org in a inconsistent state (if there are other packages, and all of that got deployed). This means development teams have to rush to get the package promoted.

Also though there is no promotion required for source or data packages, it is better to bring in feature parity for all packages

aly76:
**Breaking change: affects deployment of existing artifacts to prod**
- All packages must now be promoted to be deployed to production, when using the orchestrator
- Fix getOrgDetails interface

This change is a preventative measure to avoid deploying non-release candidates to production.

### DIFF
--- a/packages/core/src/PackageMetadata.ts
+++ b/packages/core/src/PackageMetadata.ts
@@ -18,6 +18,7 @@ export default interface PackageMetadata {
     apexTestClassses?:string[];
     isTriggerAllTests?:boolean;
     isProfilesFound?:boolean;
+    isPromoted?: boolean;
     tag?:string;
     isDependencyValidated?:boolean;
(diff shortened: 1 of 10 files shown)


azlam-abdulsalam:
@aly76  can we add a temporary hidden flag 'allowNonPromotedPackages' to ease during transition phase, as there would be many packages that wont have the isPromotedFlag

azlam-abdulsalam (approved):
All good!

PR #441, opened 2021-03-16
Remove usage of npx to trigger sfdx commands

azlam-abdulsalam:
PR to remove usage of npx commands.
A relic of the past where sfdx plugin was not able to be installed in Azure Pipelines with the right permission﻿


aly76 (approved):
LGTM

PR #442, opened 2021-03-16
Check whether all package types have been promoted before publishing

aly76:
* Check whether all package types have been promoted before publishing
* Deprecate `--devhubalias` flag, as it is no longer required to check the promotion status of packages
* Add default value to `--publishpromotedonly`


azlam-abdulsalam (approved):
okie

PR #445, opened 2021-03-17
Add --npmrcpath parameter

### LINKED ISSUE #418: Add an option to create npm compatible packages.

aly76:
- Add `--npmrcpath` parameter 
- Make `--scope` parameter required
- Fix artifact name capitalisation in Prepare for NPM
- Parse `--scope parameter` 
  - Remove '@' symbols and convert to lower case

### DIFF
--- a/packages/sfpowerscripts-cli/src/impl/prepare/PrepareImpl.ts
+++ b/packages/sfpowerscripts-cli/src/impl/prepare/PrepareImpl.ts
@@ -8,6 +8,7 @@ import { SfdxApi } from "../pool/sfdxnode/types";
 import PrepareASingleOrgImpl, {
   ScriptExecutionResult,
 } from "./PrepareASingleOrgImpl";
+import path = require("path");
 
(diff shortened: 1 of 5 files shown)


packages/sfpowerscripts-cli/src/impl/prepare/PrepareImpl.ts
```diff
@@ -181,6 +187,18 @@ export default class PrepareImpl {
...
     if (this._isNpm) {
+      if (this._npmrcPath) {
+        fs.copyFileSync(
+          this._npmrcPath,
+          path.resolve(".npmrc")
+        );
+
+        if (!fs.existsSync("package.json")) {
+          // package json is required in the same directory as .npmrc
+          fs.writeFileSync("package.json", "{}");
```
azlam-abdulsalam, line 198: Doesnt it  need atleast 1 attribute?

azlam-abdulsalam (changes requested):

PR #449, opened 2021-03-18
Fix deployment of scoped artifacts 

### LINKED ISSUE #418: Add an option to create npm compatible packages.

aly76:
- Fix deployment of scoped artifacts
  - Remove hat ^ to match scopes  
  - Add case-insensitivity to accommodate NPM packages
- Fix regexp in ArtifactFilePathFetcher


azlam-abdulsalam (approved):
All good!

PR #454, opened 2021-03-19
Validate package names during build

### LINKED ISSUE #418: Add an option to create npm compatible packages.

aly76:
Package names should comply with NPM naming convention, with the exception of uppercase letters which are permitted.


azlam-abdulsalam (approved):
okie

PR #458, opened 2021-03-22
Refactor FetchImpl

aly76:
Move `ReleaseDefinition` and `validateReleaseDefinition()` out of FetchImpl, so that they can be used by the `orchestrator:release` command as well.


packages/sfpowerscripts-cli/src/impl/release/validateReleaseDefinition.ts
```diff
@@ -0,0 +1,53 @@
+import ReleaseDefinition from "../release/ReleaseDefinitionInterface";
+const Validator = require('jsonschema').Validator;
+
+export default function validateReleaseDefinition(
```
azlam-abdulsalam, line 4: can we have unit tests for this?

PR #469, opened 2021-03-31
Fix OrgCoverage Packages should not be tested for coverage check

azlam-abdulsalam:
This PR fixes a bug where it checks for coverage incorrectly for org dependent package﻿


aly76 (approved):
LGTM

PR #481, opened 2021-04-07
Fix InstalledArtifactsFetcher for use case with multiple orgs

aly76:
- In cases such as `prepare`, multiple orgs are being provisioned at the same time, so each org needs to have its query result for installed artifacts stored.
- Change artifacts property to map of usernames to artifacts


azlam-abdulsalam (approved):
okie

azlam-abdulsalam:
@aly76  All good, lets get a hotfix!

PR #493, opened 2021-04-19
Create 004-Identifying-packages-in-npm.md

azlam-abdulsalam:
I am not sure whether I am overcooking this one. Early draft for a mechanism to identify packages when pushed to a single package registry like the Github Packages


LawrenceLoz:
This will be really helpful to introduce and option in decision sounds great. FWIW from my perspective having packages reflecting branch names in registry has some benefits - branches evolve in parallel and it's helpful to see the sequence of version numbers separated by branch (I'm using Github packages, in case that's relevant to UI considerations)

azlam-abdulsalam:
The updated  release changelog feature displays package, orgs and work items all filtered by the branch the release is activated, hence making it easy to rollback. This proposal makes it complicated amd results in rebuilding artifacts which is not ideal.

PR #500, opened 2021-04-20
Creating release branches

aly76:
# Cutting release branch does not create a set of rc artifacts

* Status: Proposed  

When cutting release branches, a set of rc artifacts are not automatically created. The user must disable `--diffcheck` to generate a set of rc artifacts, or manually publish the last dev artifacts to the rc feed, or update the rc tags in the case of single registry.
This process might be made simpler for the end-user, so that intervention is not required when cutting a release branch.

## Options
1. **CLI helper tools that automates release cutoff tasks**
    - Create a CLI tool that automates the tasks required when cutting a release branch: fetch last set of artifacts from develop feed and publish them to the rc feed, or updates the rc NPM tags to point to the last set of develop artifacts.
    - The CLI would have to authenticate to both the dev and rc feeds, which could be achieved using two separate .npmrc files
2. **Automatically create new rc artifacts for a new branch**
    - The build command will automatically create rc artifacts on a release cutoff, instead of showing that there is nothing to be built
    - This can be achieved by encoding the branch in the build metadata, so that tags can be further filtered by branch e.g. major.minor.patch-buildNumber+<branch>
3. **Disable `--diffcheck` when cutting new release branches**
    - Lowest effort
    - User is responsible for toggling `--diffcheck` when cutting release branches, so that a set of rc artifacts is created

## Decision


LawrenceLoz:
Great to see this this is an area being looked at. I was originally planning to use release-specific branches for our workflow but moved to a persistent UAT/pre-prod branch instead as this seemed to work more effectively with the tag filtering design used with --diffcheck. It will be great to support both models in future and this would help also with transitioning or splitting pipelines to new branches.

Under option 2, am I right in thinking would this would always result in a complete rebuild of all packages even if no difference? It would be great if a partial release involving subset of package builds could be supported simply under any approach (it sounds like this would work under option 1)

azlam-abdulsalam:
@LawrenceLoz  Great to see that you figure it out, it is a bit tough one and missing in docs.

azlam-abdulsalam:
@aly76  This proposal need to be reworked.

As per our last discussion on this issue https://github.com/Accenture/sfpowerscripts/issues/495 , the current functionality is ideal. We should move away from recommending two feeds at all,  projects should be using a single artifact feed per repo. We could have a helper  tag packages in bulk, more from a usability enhancement perspective.(but not a priority)

On the release definition side, we leave it to the user to use LATEST_TAG or exact versions or npm tag as there is no need to rebuild an existing artifact unless there is a change
 
@LawrenceLoz fyi

PR #509, opened 2021-04-28
Add support for checkpoints in prepare

azlam-abdulsalam:
We have noticed there is an amount of flakiness in scratchOrgs, when it comes to settings. Some of the settings doesnt get enabled which results in failures.

 
This becomes an issue when the prepare command is used with --succeedondeploymenterrors with the intent to get a partially completed scratch org. The scratch orgs often returned are plain empty (only has dependencies installed). 

To counter this, this PR introduces a concept of 'checkpointForPrepare'. If at least one checkpoint has been satisfied (deployed to the SO), only then is the scratch org persisted in the pool. This reduces those near empty scratch orgs which are returned during validate, which will skew the PR Times﻿


aly76 (approved):
LGTM

PR #512, opened 2021-05-05
Suppress stdout when publishing to NPM

aly76:
- Suppress stdout when publishing to NPM
- Minor refactor to functions


azlam-abdulsalam (approved):
okie

PR #513, opened 2021-05-06
Publish artifacts with source branch as the NPM tag

### LINKED ISSUE #492: Publish should not  push packages with LATEST tag while publishing to NPM
**Describe the bug**
sfpowerscripts orchestrator:publish command provides options to publish artifacts to NPM. However artifacts can be generated from various branches and can be fed into the same package registry. This results in LATEST tag switching between packages generated from develop and release and can cause an incorrect version to be deployed to an environment, for eg: in a scaled trunk based model,  packages created from the release branch is to be used only for higher environments such as production. 

**Expected behavior**
sfpowerscripts should default to a convention based approach, any packages that are generated from branches such as dev/master/main should be tagged with 'dev' and any packages that are generated from release* should be tagged as rc by default, withoptions provided to the user to change the npm, tag as they wish.

As a work around, any users who are using publish to NPM should immediately add tags based on the branch where their packages are being based on.

aly76:
Fixes #492 
- By default, publish artifacts with source branch as the NPM tag
- If provided, `--npmtag` takes precedence
- Artifact version is no longer validated in the release definition. Instead, NPM will return error for non-existent version.


azlam-abdulsalam (approved):
Okie

Issue #514, opened 2021-05-06
Generate Changelog functionality only works for self hosted agents


azlam-abdulsalam:
**Describe the bug**
The recently released release command which features the generate changelog functionality only works with self hosted agents and is not able to push the changes to cloud hosted agents

**To Reproduce**
1.  Create a release with --generatechangelog flag on a cloud hosted agent such as github action
2. Observe that the command fails to push

**Expected behavior**
The command should be able to push the changelog in all kinds of agents

https://github.com/dxatscale/easy-spaces-lwc/actions/runs/815791071

aly76:
Solutions:

- User to setup SSH agent 
  - For hosted agents, look at action/ssh-agent
  - PRO: user may already need access to multiple private repo's and thus needs to setup SSH in order to get around GitHub actions limitation to the source repo 
  - CON: changelog does not work without SSH
  - CON: Higher-activation energy required from user

- Use the authenticated source repo instead of cloning a repo to tmp dir
  - Checkout the changelog branch & return to the original detached HEAD when done
    - Need to stash CWD before changing branches to avoid losing data 
    - CON?: the changelog repoURL is no longer a user-input, and must be the same as the source repo 
    - CON: `changelog:generate` is still dependent on SSH
    - CON: Compatibility must be maintained with this `changelog:generate` resulting in two control paths  
    - PRO: works without SSH
  - Copy the CWD to a tmp directory, which maintains auth to the source repo
    - PRO: free to modify tmp repo without having to revert to the original commit once done 
    - CON: Same as above
    
- Release ccommand can take GitHub token as a flag
  - Authenticate tmp dir using passed token 
  - CON: need to get auth token from CICD, which may not always be available?
  - CON: more work for the user

PR #517, opened 2021-05-09
Support for agents authenticated to GIT using non-SSH in release changelog functionality

aly76:
#514 - Add support for agents in the `orchestrator:release` command's `--generatechangelog` functionality using any other authentication mechanism other than SSH

**breaking** `repoUrl` no longer accepted in the release definition yaml 
**breaking** `--repoUrl` no longer a flag in `changelog:generate`. The command requires the cwd to be the source directory.


azlam-abdulsalam:
changelog:generate should provide an option of no push, so the user can push it to any repo

azlam-abdulsalam (approved):

PR #528, opened 2021-05-23
feat/523 PromoteInRelease

### LINKED ISSUE #523: Promoted artifacts need to be republished
sfpowerscripts in its current design, implies that promoted packages have to be republished (as it internally changes the artifact state to mark it as promoted). This would only work when the user is having two artifact registries.If its a single artifact repo such as npm is used, the package gets rejected, if its already published (say from trunk). This implies that the packages have to be rebuilt when a release branch is cut.

To enable it the following should be done
- The promote flag should be removed from the internal state
- Promotion should be moved into the release definition where packages are promoted  before deploying to a particular environment

azlam-abdulsalam:
Fixes #523

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ lerna-debug.log
 
 packages/core/coverage
 packages/sfpowerscripts-cli/coverage
+ .DS_Store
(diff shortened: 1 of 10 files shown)


aly76 (commented):
LGTM

aly76 (approved):
Looks good

PR #529, opened 2021-05-25
Remove automated branch generation for changelogs

azlam-abdulsalam:
The last version introduce a feature where changelog used to generate the changelog branch by observing the branch which the artifact is developed from.
This resulted in a bug, which prevents scenario where a single artifact repository is being used along with a scaled trunk model where the packages are
not recreated, resulting in incorrect changelog

To fix these scenarios, we are bringing back the branch flag back and let the users push the changelog to any preferred branch, however one that track 
the trunk

The change implies  there needs to be a changelog that tracks a branch where the package is originated

Examples:
   main --> "changelog-main" (during normal development)
   release/1.0 -> "changelog-release_1.0" (when cut a release branch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ lerna-debug.log
 
 packages/core/coverage
 packages/sfpowerscripts-cli/coverage
+.DS_Store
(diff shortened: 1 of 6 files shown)


aly76 (approved):
Looks good

PR #530, opened 2021-05-27
Revert --publishpromotedonly to only check unlocked packages

aly76:
- Un-deprecate the devhub alias flag 
- Only check whether unlocked packages have been promoted 

Addendum to #528


azlam-abdulsalam (approved):
okie

PR #539, opened 2021-06-01
Ignore aliasified packages automatically on validate and prepare stages

### LINKED ISSUE #444: An aliasified package should be auto ignored in Validate stage or alternate mechanism to be provided
An aliasified source package at the moment, should explicitly be ignored in validate stage. There is no documentation clarifying this nor any alternate approach is proposed on how an aliasified source package should be pushed to scratch org

aly76:
Fixes #444


azlam-abdulsalam (approved):
Okie

PR #542, opened 2021-06-03
 Encapsulate child process command parameters with double quotes

aly76:
Encapsulating parameters in double quotes prevents them from being parsed
incorrectly when a parameter contains spaces.

Added double quotes to parameters:
- target org
- devhub username
- file paths
- package names
- tags
- test names
- permission sets

Added single quote to commands that accept a string containing key/value pairs
as a parameter e.g. force:data:record:create

### DIFF
--- a/packages/core/src/artifacts/ArtifactInstallationStatusUpdater.ts
+++ b/packages/core/src/artifacts/ArtifactInstallationStatusUpdater.ts
@@ -52,12 +52,12 @@ export default class ArtifactInstallationStatusUpdater {
         SFPLogger.log("Updating Org with new Artifacts "+packageName+" "+packageMetadata.package_version_number+" "+(artifactId?artifactId:""), null, packageLogger, LoggerLevel.INFO);
         if (artifactId == null) {
           cmdOutput = child_process.execSync(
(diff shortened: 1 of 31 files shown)


azlam-abdulsalam:
should we be encapsulating or let the user do it? I am worried on what the behaviour if i pass in with quotes as expected from command line tools

aly76:
> should we be encapsulating or let the user do it? I am worried on what the behaviour if i pass in with quotes as expected from command line tools

There's two points at which flags get parsed: once when running the sfpowerscripts command on the terminal and again when executing a sfdx command in a child process. Any quotes passed on the terminal won't be retained once they get assigned to flags object as a string. The only way that quotes could be part of the string is if the user passed in the flag as `--targetorg "\"dev hub\""`

rody:
@aly76 could we use `spawn` or `spawnSync` instead of `exec` with the `shell` option set to `false`? This way we would bypass the quote processing by the shell when calling the child process.

aly76:
> @aly76 could we use `spawn` or `spawnSync` instead of `exec` with the `shell` option set to `false`? This way we would bypass the quote processing by the shell when calling the child process.

Thanks for the heads up @rody. We will keep this in mind when resolving this issue.

We are now trying to avoid running SFDX commands in a child process as much as possible, and hitting the JSForce API's directly.

azlam-abdulsalam:
I am closing this one to be redone using spawn or spawnSync

PR #546, opened 2021-06-07
Refactor Logger

### LINKED ISSUE #522: Revamp Logger
The current logger should be replaced with a proper logging library that supports multiple   scenarios, such as extended logging to a file for orchestrator commands, where as displaying only the important info on the terminal

azlam-abdulsalam:
Closes #522  Revamp existing SFPLogger

This cleans up the existing SFPLogger implementation to support chalk and allow more flexible signature.


aly76 (approved):
Okay

PR #547, opened 2021-06-08
Update Sfpowerscripts Artifact package version ID in Prepare

aly76:


azlam-abdulsalam:
@aly76 I dont think we should not update the develop branch with an early update on the package version id, as this feature is not officially released? Can we add a caveat stating if you are using alpha, use this new package?

aly76:
> @aly76 I dont think we should not update the develop branch with an early update on the package version id, as this feature is not officially released? Can we add a caveat stating if you are using alpha, use this new package?

Okay, I'll just put the docs in a separate open PR that we can merge before going to prod

azlam-abdulsalam (approved):
okie

PR #548, opened 2021-06-08
Update docs with new Sfpowerscripts Artifact package version ID

### LINKED ISSUE #476: Migrate sfpowerscripts artifact record to use custom settings
sfpowerscripts artifacts is an unlocked package that keeps track of all the packages installed in the org.  The current implementation utilizes a custom object to store the information. However, this has a downside, that when an org gets refreshed, the entire data gets wiped out and the traceability is lost. 

To solve this, these records should be stored in custom settings, which are preserved during refreshes.  The cli should support utilising custom settings to store these records with a backward compatibility layer,  so that none of the data gets lost while the migration is underway

aly76:
**To be merged before releasing to prod**


azlam-abdulsalam (approved):
Approved

PR #555, opened 2021-06-14
Query SfpowerscriptsArtifact custom object using API

aly76:
- Query SfpowerscriptsArtifact custom object using API, in ArtifactMigrator
- Run queries again if either failed
- Show SfpowerscriptsArtifact__c deprecation notice only if it's installed
- Add dependency `@salesforce/core` in core

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -30,7 +30,6 @@
     "ignore": "^5.1.6",
     "lodash": "^4.17.21",
     "rimraf": "^3.0.2",
-    "salesforce-alm": "^52.0.1",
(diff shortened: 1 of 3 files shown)


packages/core/package.json
```diff
@@ -29,6 +30,7 @@
     "ignore": "^5.1.6",
     "lodash": "^4.17.21",
     "rimraf": "^3.0.2",
+    "salesforce-alm": "^52.0.1",
```
azlam-abdulsalam: Do we need ALM package?

PR #580, opened 2021-07-07
Deprecate JWT Based Authentication in Validate

azlam-abdulsalam:
JWT based authentication is being deprecated in favour of authentication mechanisms that provide an access/refresh token, to align with convergence of dev and ci pools.

Introduced a flag in FetchPoolImpl constructor, so that it only fetches scratchorgs with a valid authURL in case if the pool was created using JWT.

### DIFF
--- a/packages/sfpowerscripts-cli/src/commands/sfpowerscripts/pool/fetch.ts
+++ b/packages/sfpowerscripts-cli/src/commands/sfpowerscripts/pool/fetch.ts
@@ -81,7 +81,8 @@ export default class Fetch extends SfdxCommand {
     let fetchImpl = new PoolFetchImpl(
       this.hubOrg,
       this.flags.tag,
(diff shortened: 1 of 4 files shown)


aly76 (approved):

packages/sfpowerscripts-cli/src/commands/sfpowerscripts/pool/fetch.ts
```diff
@@ -81,7 +81,8 @@ export default class Fetch extends SfdxCommand {
     let fetchImpl = new PoolFetchImpl(
       this.hubOrg,
       this.flags.tag,
-      null,
+      false,
+      false,
```
aly76, line 85: authURLEnabledScratchOrg true?

aly76 (approved):

PR #582, opened 2021-07-07
Prepare to Enforce Non JWT  Auth Mechanisms

azlam-abdulsalam:
- Add Validation on Prepare to enforce pools are created with non JWT
- Remove unused import

### DIFF
--- a/packages/sfpowerscripts-cli/src/impl/prepare/PrepareImpl.ts
+++ b/packages/sfpowerscripts-cli/src/impl/prepare/PrepareImpl.ts
@@ -3,8 +3,8 @@ import { Org } from "@salesforce/core";
 import { PoolConfig } from "../pool/PoolConfig";
 import { PreparePoolInterface } from "./PreparePoolInterface";
 import PreparePool from "./PreparePool";
-
-
+import OrgDisplayImpl from "@dxatscale/sfpowerscripts.core/lib/sfdxwrappers/OrgDisplayImpl";
+import isValidSfdxAuthUrl from "../pool/prequisitecheck/IsValidSfdxAuthUrl";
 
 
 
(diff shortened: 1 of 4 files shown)


packages/sfpowerscripts-cli/src/impl/prepare/PrepareImpl.ts
```diff
@@ -25,9 +26,14 @@ export default class PrepareImpl {
 
   public async exec()
   {
-
     let poolPreparer:PreparePoolInterface;
 
+    let orgDisplayImpl = new OrgDisplayImpl(null, this.hubOrg.getUsername());
+    let orgDisplayResult = await orgDisplayImpl.exec(true);
+
+    if(!(orgDisplayResult.sfdxAuthUrl && isValidSfdxAuthUrl(orgDisplayResult.sfdxAuthUrl)))
```
aly76, line 34: See changes

aly76 (approved):

Vacster:
@azlam-abdulsalam could I get some context on why this was required?

I don't understand why preparing a pool of scratch orgs cannot be done with a JWT Authentication and I can't find the reasoning documented anywhere.

--- 
Moreover, the checks added explicitly make it so we need an SFDX auth URL. This is weird because that requires a refreshToken on it.

From the two ways that Salesforce [recommends doing Server-to-Server OAuth Authorisation](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_flows.htm&type=5) they both explicitly don't include refreshTokens:
- [JWT](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_jwt_flow.htm&type=5#:~:text=NOTE%20This%20flow%20never%20issues%20a%20refresh%20token.)
- [Client Credentials](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_client_credentials_flow.htm&type=5#:~:text=NOTE%20This%20flow%20doesn%E2%80%99t%20support%20refresh%20tokens.)

I find it strange that, if we're supposed to use the `prepare` command in a CI environment, we can't use Salesforce's recommended OAuth flows to actually connect to the DevHub. What OAuth method are we supposed to be using in a CI environment if not either of these two?

Vacster:
@aly76 as you approved this, do you perhaps know the answer to my previous comment?
https://github.com/dxatscale/sfpowerscripts/pull/582#issuecomment-1739546545

PR #584, opened 2021-07-08
Add a noColor option to disable colors

### LINKED ISSUE #475: Add Chalk Support
Most CI/CD supports  ansi based colour schemes. As sfpowerscripts logs are often lengthy,  the readability is impacted. Having ansi based color coding would help.

azlam-abdulsalam:
sfpowerscripts by default will enforce colors in all logs. This include terminal emulation in CI runners
like GitHub Actions or Azure Pipelines. This is for CI systems which doesnt support colors and result in
logs being garbled. By adding a`SFPOWERSCRIPTS_NOCOLOR` environment variable, users can turn off logs being displayed with colors in such systems  

 - Disable colors if `SFPOWERSCRIPTS_NOCOLOR` environment variable is true


packages/core/src/logger/SFPLogger.ts
```diff
@@ -55,6 +53,12 @@ export const COLOR_KEY_VALUE=chalk.black.bold.bgGreenBright;
 export default class SFPLogger {
   public static logLevel: LoggerLevel = LoggerLevel.INFO;
 
+  static setColorLevel(nocolor:boolean) {
+  if(nocolor)
+    chalk.level = 0
+  else 
+    chalk.level = 1;
```
aly76: By default, Chalk will automatically detect the level of colour support based on the environment

aly76 (changes requested):

packages/core/src/logger/SFPLogger.ts
aly76: Hence, we don't need an else case

aly76 (approved):

PR #598, opened 2021-07-13
Push source using force parameter

aly76:
When deploying metadata duplicated across packages, a force push is required to override conflicts.


azlam-abdulsalam (approved):
okie

PR #606, opened 2021-07-17
Refactor PackageTestCoverage

aly76:
Move ApexClass, ApexTrigger and ApexCodeCoverageAggregate queries into own
class.

Add test case for calculating package coverage with untouched classes and
triggers.


azlam-abdulsalam (approved):
okie

PR #622, opened 2021-07-26
Refactor Sfdmu execution

azlam-abdulsalam:
- Refactor API Signature
- Add Support for Logging in Execute Command
- Fix Signature
- Move Sfdmu command invocation to sfdmu run impl

### DIFF
--- a/packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
+++ b/packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
@@ -1,7 +1,5 @@
 import PackageMetadata from "../../PackageMetadata";
-import child_process = require("child_process");
-import { onExit } from "../../utils/OnExit";
-import fs = require("fs");
+import fs = require("fs-extra");
 import ArtifactInstallationStatusChecker from "../../artifacts/ArtifactInstallationStatusChecker";
 import {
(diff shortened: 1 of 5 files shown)


packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
```diff
@@ -209,11 +208,27 @@ export default class InstallDataPackageImpl {
...
 
-    SFPLogger.log(`Generated Command ${command}`, null, this.packageLogger);
-    return command;
+    if (fs.pathExistsSync(path.join(packageDirectory, "export.json"))) {
+      SFPLogger.log(
+        `Found export.json in ${packageDirectory}.. Utilizing it as data package and will be deployed using sfdmu`,
+        LoggerLevel.INFO,
+        this.packageLogger
+      );
+      return "sfdmu";
```
aly76, line 219: Perhaps we can use an enumerator for the type of data package

aly76 (approved):

packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
azlam-abdulsalam: True.. will be adding that as part of the next PR

PR #627, opened 2021-07-28
Unzip artifact for each scratch org

aly76:
Resolves concurrency problem when preparing scratch orgs, where the same data
package directory is used across installations.


azlam-abdulsalam (dismissed):
okie

PR #631, opened 2021-07-28
Fix OrgDetailsFetcher

aly76:
Set sfdxAuthUrl to undefined if unavailable, as in the case of JWT authenticated org


azlam-abdulsalam (dismissed):
okie, a comment would be nice in top of that error ..or a log inside the error set to trace

PR #634, opened 2021-07-31
Abandon scratch orgs with status of 'Deleted'

aly76:
- Do not commit scratch orgs with status of 'Deleted' to the pool
- Add scratch org info status field to org details


azlam-abdulsalam (approved):
Okie

PR #641, opened 2021-08-06
Documentation/dr dependency manager

Caitlyn-Mills:
# Dependency Manager 
•	Status: Proposed
•	Issue: #638

As the number of packages increases in large scale projects, with multiple dependencies, updating package versions and dependencies manually is time consuming and takes great care to update dependencies one by one. 

Other solutions encountered were local bash scripts as documented in both the issue https://github.com/Accenture/sfpowerscripts/issues/638 raised and in the discussions has here: https://github.com/Accenture/sfpowerscripts/discussions/468 

## Solution
The dependency manager should utilise the SemVer incrementation style https://semver.org/ and be of a similar method to the ‘lerna version’ command. It should prompt the user for the version either for individual packages or for all packages in the sfdx-project.json: (major, minor, patch, custom). 

Since salesforce does not necessarily follow the semver versioning, 
We will also not be supporting pre-release versioning at this stage. 

Steps to the command:
For each package in the sfdx-project.json the command will ask 
1. if the package version should be updated
2. If the package is a source/data package with a non-zero build number, does the build number need to be reset to 0 or stay as is 
3. if the package version has been updated, would the user like all packages with this dependency updates as well? 

The packageDirectories portion will then be updated with the input given and written to the sfdx-project.json in the root folder with the option to commit the file (but not push to the repo). 

## Decision 
A new command would be built with the above solution and will be added to sfpowerscripts


azlam-abdulsalam:
@Caitlyn-Mills  Can you add also the following

1) other solution that you  encountered
2) As salesforce packages not necessarily follow semver , how do we intend to solve this. probably only for major.minor.patch and ignore build number
3) Not clear on how the dependency are updated. Whats the default behaviour?  What if I don't want to update dependency?

Caitlyn-Mills:
@azlam-abdulsalam updated

decision records/003-Depenedency-Manager.md
```diff
@@ -0,0 +1,25 @@
...
+The dependency manager should utilise the SemVer incrementation style https://semver.org/ and be of a similar method to the ‘lerna version’ command. It should prompt the user for the version: (major, minor, patch, custom). 
+Since salesforce does not necissarily follow the semver versioning, we will provide a summary to the bump. 
+Eg. Option 1 bump Major - includes incompatible API changes
+We will also not be supporting pre-release versioning at this stage. 
+
+Steps to the command:
+For each package in the sfdx-project.json the command will ask 
+1. if the package version should be updated
+2. if the package version has been updated, would the user like all packages with this dependency updates as well? 
+3. Would the user like to update the dependencies of this package? (regardless of whether the package version was updated)
```
azlam-abdulsalam: I didn't understand this one? Why do we need to do this?

decision records/003-Depenedency-Manager.md
```diff
@@ -0,0 +1,25 @@
...
+•	Status: Proposed
+•	Issue: #638
+
+## Context and Problem Statement 
+As the number of packages increases in large scale projects, with multiple dependencies, updating package versions and dependencies manually is time consuming and takes great care to update dependencies one by one. 
+
+Other solutions encountered were local bash scripts as documented in both the issue https://github.com/Accenture/sfpowerscripts/issues/638 raised and in the discussions has here: https://github.com/Accenture/sfpowerscripts/discussions/468 
+
+## Solution
+The dependency manager should utilise the SemVer incrementation style https://semver.org/ and be of a similar method to the ‘lerna version’ command. It should prompt the user for the version: (major, minor, patch, custom).
```
azlam-abdulsalam: Lerna has a no prompt support as well, where we say Major it updates major for all packages

decision records/003-Depenedency-Manager.md
```diff
@@ -0,0 +1,25 @@
...
+## Context and Problem Statement 
+As the number of packages increases in large scale projects, with multiple dependencies, updating package versions and dependencies manually is time consuming and takes great care to update dependencies one by one. 
+
+Other solutions encountered were local bash scripts as documented in both the issue https://github.com/Accenture/sfpowerscripts/issues/638 raised and in the discussions has here: https://github.com/Accenture/sfpowerscripts/discussions/468 
+
+## Solution
+The dependency manager should utilise the SemVer incrementation style https://semver.org/ and be of a similar method to the ‘lerna version’ command. It should prompt the user for the version: (major, minor, patch, custom). 
+Since salesforce does not necissarily follow the semver versioning, we will provide a summary to the bump. 
+Eg. Option 1 bump Major - includes incompatible API changes
+We will also not be supporting pre-release versioning at this stage.
```
azlam-abdulsalam, line 14: Can you elaborate here? What does pre-release mean?

azlam-abdulsalam (changes requested):
Few lines need some more clarity.

decision records/003-Depenedency-Manager.md
Caitlyn-Mills: A pre-release version MAY be denoted by appending a dash and a series of dot separated identifiers immediately following the patch version. Identifiers MUST be comprised of only ASCII alphanumerics and dash [0-9A-Za-z-]. Pre-release versions satisfy but have a lower precedence than the associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.

decision records/003-Depenedency-Manager.md
Caitlyn-Mills: Updated

decision records/003-Depenedency-Manager.md
azlam-abdulsalam: yeah we dont need pre-release support, as salesforce packages dont support it. We need to find a way for build number for source and data packages, say if the user has any non zero build number, do we bring back to zero?

Caitlyn-Mills: We should check if the number is non-zero and ask if they want it reset to zero if it is

azlam-abdulsalam: can you add these to the document?

azlam-abdulsalam (dismissed):
I am happy with it 

@aly76  Please let me know if you have any say
@LawrenceLoz  @Dieffrei  this is what we are building up. Any suggestions?

decision records/003-Depenedency-Manager.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Context and Problem Statement 
+As the number of packages increases in large scale projects, with multiple dependencies, updating package versions and dependencies manually is time consuming and takes great care to update dependencies one by one. 
+
+Other solutions encountered were local bash scripts as documented in both the issue https://github.com/Accenture/sfpowerscripts/issues/638 raised and in the discussions has here: https://github.com/Accenture/sfpowerscripts/discussions/468 
+
+## Solution
+The dependency manager should utilise the SemVer incrementation style https://semver.org/ and be of a similar method to the ‘lerna version’ command. It should prompt the user for the version either for individual packages or for all packages in the sfdx-project.json: (major, minor, patch, custom). 
+
+Since salesforce does not necissarily follow the semver versioning, we will provide a summary to the bump.
```
azlam-abdulsalam:
```suggestion
Since salesforce does not necessarily follow the semver versioning, 
```

decision records/003-Depenedency-Manager.md
```diff
@@ -0,0 +1,26 @@
...
+## Context and Problem Statement 
+As the number of packages increases in large scale projects, with multiple dependencies, updating package versions and dependencies manually is time consuming and takes great care to update dependencies one by one. 
+
+Other solutions encountered were local bash scripts as documented in both the issue https://github.com/Accenture/sfpowerscripts/issues/638 raised and in the discussions has here: https://github.com/Accenture/sfpowerscripts/discussions/468 
+
+## Solution
+The dependency manager should utilise the SemVer incrementation style https://semver.org/ and be of a similar method to the ‘lerna version’ command. It should prompt the user for the version either for individual packages or for all packages in the sfdx-project.json: (major, minor, patch, custom). 
+
+Since salesforce does not necissarily follow the semver versioning, we will provide a summary to the bump. 
+Eg. Option 1 bump Major - includes incompatible API changes
```
azlam-abdulsalam:
```suggestion

```

dieffrei:
@azlam-abdulsalam it is related to that helper command to update? Or are we going to have it automatically for package builds?
A pain point is how to make sure that my salesforce package version is compatible with npm.
Today I have the feedback that code works fine, but I can not have the feedback if that package version (unlocked/source/data) if they are publishable in NPM (We can not update artifacts)

azlam-abdulsalam:
It is a helper command to update package version numbers.

I didn't understand the  secondary ask about publishing to npm. We dont see any issue as long as you are passing the unique and incrementing build number to the build command

azlam-abdulsalam (dismissed):

decision records/003-Depenedency-Manager.md
```diff
@@ -0,0 +1,25 @@
...
+Steps to the command:
+For each package in the sfdx-project.json the command will ask 
+1. if the package version should be updated
+2. If the package is a source/data package with a non-zero build number, does the build number need to be reset to 0 or stay as is 
+3. if the package version has been updated, would the user like all packages with this dependency updates as well? 
+
+The packageDirectories portion will then be updated with the input given and written to the sfdx-project.json in the root folder with the option to commit the file (but not push to the repo). 
+
+## Decision 
+Development to commence on the solution and be included in the september release
```
azlam-abdulsalam:
```suggestion
A new command would be built with the above solution and will be added to sfpowerscripts
```

LawrenceLoz:
Hi @azlam-abdulsalam and @Caitlyn-Mills 

If I'm understanding right, the proposal is to have the helper command iteratively ask the user whether to update version for all packages in the JSON, with an option of whether to cascade this new version to dependencies in other packages.

My feeling is that as a person coordinating a release I may still have some challenges:

1. I might not know which packages had components changed, and so might miss a package which should be incremented or increment one unnecessarily
2. If I choose the option to update versions in all dependent packages, then a build command using diffcheck and referencing the new JSON would rebuild these packages regardless of whether the contents of the packages themselves has changed. This might be useful sometimes but I think most of the time we wouldn't want to build and deploy new versions when package contents haven't changed. Also, unless I've also incremented the version of these packages I wouldn't be able to promote these rebuilt versions if I've promoted these previously.

To add the most value and help a release manager increment only appropriate packages, I'd suggest the command should identify what's changed since the last release by comparing current package files to their state at a supplied git tag.

If done in this way, then for each package identified as changed, the user could be asked:

 - Whether the major, minor or patch number should increment (or whether no increment should be made)
 - Whether to reset the build number for source/data packages
 - Whether to cascade the latest version to dependency references in either all referencing packages, others which have changed themselves, or selected packages

This way it would be much easier to ensure only packages which need a rebuild have versions updated.

Also (appreciate this might be beyond the scope of this work item) it would be great to build this in a way that we can supply defaults for the options to run this in a non-interactive way and automate incrementing completely

azlam-abdulsalam:
@LawrenceLoz  The proposal is to help a developer increment a version number of a single package. While the developer is doing that, provide them with a guided experience on what to do with the dependent packages.

I agree with you on using our signature diff feature ;) and provide some automation to detect changed packages. I will discuss this with @Caitlyn-Mills 

For the end to end automation, as in auto update of version in the pipeline, I am leaning towards using conventional commits, so if the commit - message contains the corresponding messaging, we auto increment.

@Caitlyn-Mills @aly76

LawrenceLoz:
Thanks for the details @azlam-abdulsalam, sounds good and would be great to see the diff being utilised if it feels this fits with the objective. And commit messages would be a nice way of passing increment instructions in a way which is shown clearly in the history

PR #652, opened 2021-08-13
Add sample how to pass build number to build command

dieffrei:


dieffrei:
@azlam-abdulsalam I just add a really small sample just to make clear how to pass the build number to quick build command

azlam-abdulsalam:
Thanks @dieffrei  can you clarifiy that this is for Gitlab, as each CI/CD system  treats that number differently

azlam-abdulsalam (approved):
okie

PR #654, opened 2021-08-13
Add pool:metrics:publish command

### LINKED ISSUE #647: Pool Metrics Reporter
As pools are currently being fetched by the user directly using the CLI, there is no mechanism for it to report the number of remaining scratch org's (earlier when using fetch in the pipeline)

Hence, we need a command that will be run on a CI/CD system in X mins interval (depending on the schedule in the pipeline) and report the metrics of all remaining scratch orgs across all pools in the devhub

aly76:
Publish metrics about scratch org pools to your observability platform, via StatsD or direct APIs for supported platforms.

closes #647


azlam-abdulsalam (approved):
okie

PR #655, opened 2021-08-17
Fix logging for package:<type>:create commands

### LINKED ISSUE #626: Improve sfpowerscripts:package:unlocked:create output
**Describe the bug**
Current logs of sfpowerscripts:package:unlocked:create are not user-friendly and should be updated to match a similar output as in orchestrator build commands 

**To Reproduce**
Current output produced 
image

**Expected behavior**
Output which could/should be produced 
image

aly76:
Use console logger for individual package creation commands, instead of logging to file


azlam-abdulsalam (approved):
okie

azlam-abdulsalam:
Thanks @aly76  Lets release this ASAP!

PR #658, opened 2021-08-18
Update PMD analysis to add pmd minimumpriority option

DanielCalle:
I have added the option of minimumpriority option to PMD analizer. It is really useful to be able to apply the rules that cause an error and those that we can see as warnings without separating in different ruleset.


azlam-abdulsalam:
@aly76  Can you check whether this would fix https://github.com/Accenture/sfpowerscripts/issues/432

aly76 (dismissed):

azlam-abdulsalam:
@DanielCalle  we will be revising this in a different model. We would rather be displaying all errors, and providing an option to determine at what level it should break .

azlam-abdulsalam:
@DanielCalle  We have introduced an alternate in #667

PR #659, opened 2021-08-18
Update PMD analysis to apply rulesetpath option

DanielCalle:
I could see in the code that the rulesetpath option could be used to pmd, but this was not actually applied. I have added the necessary changes to be able to use this feature already available in the sfpowerkit plugin


azlam-abdulsalam:
Thanks @DanielCalle  for this. I had removed the ruleset path as part of the hotfix earlier. Will definitely bake this in

aly76 (approved):

PR #660, opened 2021-08-19
Fix misusage of test results from previous runs

aly76:
Delete test-run-id.txt and test-result-codecoverage.json, after command
execution.


azlam-abdulsalam (approved):
okie

PR #667, opened 2021-08-26
Add priority --threshold flag to PMD command

### LINKED ISSUE #432: Analyze PMD should provide options on when to break build
Analyze PMD should have options for  break build and also for warning based on options provided by a user

aly76:
Closes #432 

When a threshold is specified, fail for violations with a priority less than or
equal to the threshold.

Refactor output of function used to parse PMD XML output file

Fix regression in `--istobreakbuild` deprecated flag

When no violations are found, show a success message

Throw an error when PMD fails to generate output, the output XML file is empty or does not belong to PMD.


azlam-abdulsalam (approved):
Good to go

PR #675, opened 2021-08-30
Feature/version manager

Caitlyn-Mills:


packages/core/src/project/Version.ts
```diff
@@ -0,0 +1,122 @@
+import semver = require("semver");
+
+/**Helper functions for updating the project config */
+
+export default class Version {
```
azlam-abdulsalam: I think this should be in Package as opposed to Project.

azlam-abdulsalam: Probably tie in with SFPPackage

azlam-abdulsalam: Should have a constructor which takes in a package and  operate on top of it

packages/core/src/project/Version.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+
+  /**
+   * 
+   * @param dependencies a map of dependencies chosen to be updated
+   * @param packages the packages and dependencies with the current versions
+   * @returns the updated packages and dependencies
+   */
+  public updateDependencies(dependencies, packages) {
```
azlam-abdulsalam: Should be part of another class

packages/sfpowerscripts-cli/src/commands/sfpowerscripts/package/version/manage.ts
```diff
@@ -0,0 +1,225 @@
...
+       * SEMVER library 
+       * Flag for build number
+       * Greenfield project
+       * Confirmation of updates
+       * 
+       * Separating into modular functions in core library 
+       */
+
+
+      let version = new Version();
```
azlam-abdulsalam: Need to split it further, move the logic from the UI.. to a Impl class

packages/sfpowerscripts-cli/src/commands/sfpowerscripts/package/version/manage.ts
```diff
@@ -0,0 +1,225 @@
...
+        //First iteration of packages to update version numbers 
+        for (let pkg of packages) {
+          console.log(`${pkg.package} version number: ${pkg.versionNumber}`)
+          let newMajor = version.getMajor(pkg.versionNumber);
+          let newMinor = version.getMinor(pkg.versionNumber);
+          let patch = version.getPatch(pkg.versionNumber);
+          let custom = false;
+
+          let skip = false;
+          await inquirer.prompt([{ type: 'list', name: 'selection', message: `Would you like to update ${pkg.package}?`, choices: ['Major: ' + newMajor, 'Minor: ' + newMinor, 'Patch: ' + patch, 'Custom', 'Skip'] }]).then((selection) => {
```
azlam-abdulsalam: Can we have inquirer prompts and response as a logical function

azlam-abdulsalam (changes requested):
I have made some comments

packages/core/src/project/Version.ts
Caitlyn-Mills: Yeah will do constructor with package input and exec can be to call the update method. But I will still leave the getter methods as public.

Caitlyn-Mills: Changes made - I think it makes it a little unclear, but changes made.

packages/sfpowerscripts-cli/src/commands/sfpowerscripts/package/version/manage.ts
```diff
@@ -0,0 +1,219 @@
...
+
+
+// Initialize Messages with the current plugin directory
+Messages.importMessagesDirectory(__dirname);
+
+// Load the specific messages for this file. Messages from @salesforce/command, @salesforce/core,
+// or any library that is using the messages framework can also be loaded this way.
+const messages = Messages.loadMessages('@dxatscale/sfpowerscripts', 'manage_versions');
+
+export default class ManageVersions extends SfpowerscriptsCommand {
```
azlam-abdulsalam: should be renamed to "manage"

packages/core/src/project/Version.ts
Caitlyn-Mills: Done

packages/sfpowerscripts-cli/src/commands/sfpowerscripts/package/version/manage.ts
Caitlyn-Mills: Done

aly76:
This feature is being moved to the sfp-cli

packages/sfpowerscripts-cli/src/impl/package/VersionManageImpl.ts
```diff
@@ -0,0 +1,248 @@
...
+    protected noPrompt;
+    protected resetBuildNumber;
+    protected allDependentPkgs = [];
+
+
+    constructor(projectConfig, allpackages, dependents, packageFlag, version, noprompt, resetbuildnumber) {
+
+        this.projectConfig = projectConfig;
+        this.packageFlag = packageFlag;
+        this.allPackages = allpackages;
```
aly76, line 24: The `allPackages` prop is not required. Impl should just receive a list of packages from the caller.

packages/sfpowerscripts-cli/src/impl/package/VersionManageImpl.ts
```diff
@@ -0,0 +1,248 @@
...
+import ProjectConfig from "@dxatscale/sfpowerscripts.core/lib/project/ProjectConfig"
+
+export default class VersionManageImpl {
+
+
+    protected projectConfig;
+    protected packageFlag;
+    protected packages;
+    protected allPackages;
+    protected dependents;
```
aly76, line 13: While `dependents` or `updatedependents` is an okay name for the CLI flag, which is seen by the end-user, the implementation class' constructor param should be named differently e.g. `isUpdateDependents` so that it is more readable for the programmer.

Also, the properties should be statically typed.

azlam-abdulsalam:
Closing this,.as this will be part of upcoming sfp-cli

PR #711, opened 2021-09-28
Set Git username and email using environment variable

### LINKED ISSUE #710: Hard-coded user.email value in Changelog code causes git push failure
**Describe the bug**
Using sfpowerscript in a GitHub + Jenkins environment fails to push to the "changelog" when using _generatechangelog_

 I can release except error w/ pushing to **changeLog** ; _the internal system I am on checks the user's email (as valid)_ ... When I dug a bit deeper into this I noticed  in **changelog/ChangelogImpl.ts:pushChangelogToBranch** sets the **user.name** to "_sfpowerscripts_" and  **user.email** to "_sfpowerscripts@dxscale_" ... this is causing the failure. 

Customer runs checkemail anytime a push occurs (**not possible to disable**)

**To Reproduce**
Using the latest DockerImage . Run script (bash) in Jenkins 

_sfdx sfpowerscripts:orchestrator:release .... --generatechangelog --branchname changelog_

**Expected behavior**
Changelog with pushed release information

**Screenshots**

 - OS: Mac Big Sur 11.5.2
 - Version [e.g. CLI Version  eg: 1.6.6] What's in the current Docker image 
 - Salesforce CLI(sfdx cli) Version: [7.110.0]
 - CI Platform: GitHub + Jenkins

**Additional context**
All is fine if I do not use _--generatechangelog --branchname_

aly76:
- Set Git username and email using environment variable
- Refactor child processes to Simple Git
- Upgrade Simple Git dependency
- Remove deprecated import "simple-git/promise"

Closes #710


packages/core/src/git/Git.ts
```diff
@@ -1,4 +1,4 @@
-import simplegit, { SimpleGit } from "simple-git/promise";
+import simplegit, { SimpleGit } from "simple-git";
```
azlam-abdulsalam, line 1: Why cant git identity be part of this class? Dont we need to use this in sfp-cli?

packages/sfpowerscripts-cli/src/impl/parallelBuilder/BuildImpl.ts
```diff
@@ -545,13 +540,13 @@ export default class BuildImpl {
       }
     } else {
       if (packageType === "Source" || packageType == "Unlocked") {
-        result = this.createSourcePackage(
+        result = await this.createSourcePackage(
```
azlam-abdulsalam, line 543: I don't think we should await here, it should return the promise to make it parallel

azlam-abdulsalam (commented):
Have couple of comments, Lets review it

packages/sfpowerscripts-cli/src/impl/parallelBuilder/BuildImpl.ts
aly76: The invocation of the createPackage() method by the scheduler is asynchronous, but the logic within the method should be synchronous.

packages/core/src/git/Git.ts
aly76: We can but it involves more work in adding more functions to `Git` and refactoring the cli to use the class.

PR #719, opened 2021-10-07
Support default directory for aliasified source packages

### LINKED ISSUE #715: Support for 'default'  folder in aliasify packages
Aliasified folders at the moment needs to have a folder for each environment.  Though the initial thought around it was more around accuracy and not getting a default folder deployed to any environment. 

This has proven to be cumbersome for some teams. It would be ideal for sfpowerscripts to support a default folder, where the default folder will be picked up if the enviroment folder is not available.

aly76:
- Support default directory for aliasified source packages

- Fix search directory for alias so that it points to the artifact directory

- Do not skip aliasfied data packages on validate and prepare stages, in BuildImpl

- Modify readdirRecursive so that it doesn't include the search directory in the
results, unless isAbsolute is given

- Add unit test for readdirRecursive

### DIFF
--- a/packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
+++ b/packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
@@ -46,21 +46,26 @@ export default class InstallDataPackageImpl {
       });
 
       if (packageDescriptor.aliasfy) {
-        const files = FileSystem.readdirRecursive(packageDescriptor.path, true);
+        const searchDirectory = path.join(this.sourceDirectory, packageDescriptor.path);
+        const files = FileSystem.readdirRecursive(searchDirectory, true);
 
-        packageDirectory = files.find(file =>
-          path.basename(file) === this.targetusername && fs.lstatSync(file).isDirectory()
-        )
+        let aliasDir: string;
 
-        if (!packageDirectory) {
+        aliasDir = files.find(file =>
+          path.basename(file) === this.targetusername && fs.lstatSync(path.join(searchDirectory, file)).isDirectory()
+        );
+
+        if (!aliasDir) {
           packageDirectory = files.find(file =>
-            path.basename(file) === "default" && fs.lstatSync(file).isDirectory()
-          )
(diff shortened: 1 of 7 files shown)


packages/core/src/sfpcommands/package/InstallDataPackageImpl.ts
```diff
@@ -46,21 +46,26 @@ export default class InstallDataPackageImpl {
...
           packageDirectory = files.find(file =>
-            path.basename(file) === "default" && fs.lstatSync(file).isDirectory()
-          )
+            path.basename(file) === "default" && fs.lstatSync(path.join(searchDirectory, file)).isDirectory()
+          );
         }
 
-        if (!packageDirectory) {
+        if (!aliasDir) {
           throw new Error(`Aliasfied package '${this.sfdx_package}' does not have a '${this.targetusername}'' or 'default' directory`);
```
azlam-abdulsalam: Can we change this?
```suggestion
          throw new Error(`Aliasfied package '${this.sfdx_package}' does not have an alias with '${this.targetusername}'' or 'default' directory`);
```

packages/core/src/sfpcommands/package/InstallSourcePackageImpl.ts
```diff
@@ -336,12 +337,27 @@ export default class InstallSourcePackageImpl {
...
+      )
+
+      if (!aliasDir) {
+        aliasDir = files.find(file =>
+          path.basename(file) === "default" && fs.lstatSync(path.join(searchDirectory, file)).isDirectory()
+        )
+      }
+
+      if (!aliasDir) {
+        throw new Error(`Aliasfied package '${this.sfdx_package}' does not have a '${this.targetusername}'' or 'default' directory`);
```
azlam-abdulsalam:
```suggestion
   throw new Error(`Aliasfied package '${this.sfdx_package}' does not have an alias with '${this.targetusername}'' or 'default' directory`);
```

azlam-abdulsalam (commented):
@aly76  Can you look into the few comments

PR #720, opened 2021-10-08
Add isSandbox to OrgDetails

aly76:
- In a multi-org environment, a production org may not be a Dev Hub, hence we need
an alternative way to 'isDevHub' for determining whether an org is prod.

- Use default directory for aliasfied packages, in deployments to sandboxes only


azlam-abdulsalam (approved):
LGTM

PR #722, opened 2021-10-14
Fix reference to length on type undefined

aly76:
When a source package is deployed and there is no test classes, sfpowerscripts will crash with the above error


azlam-abdulsalam (approved):
okie

PR #725, opened 2021-10-14
Do not include assigned scratch orgs in the available scratch org count

RobinWeymans:
When the prepare commands decides how many scratch orgs it will make, it has to compare the pool size (max_allocation) to the number of available active scratch orgs.

Currently the scripts does not take the allocation status into account. That's a problem because if the pool size is 10, and all 10 scratches are assigned, the script will not make any new scratch orgs. I'd expect that the pool always has 10 scratch orgs that are available to use.

### DIFF
--- a/packages/sfpowerscripts-cli/src/impl/pool/services/fetchers/ScratchOrgInfoFetcher.ts
+++ b/packages/sfpowerscripts-cli/src/impl/pool/services/fetchers/ScratchOrgInfoFetcher.ts
@@ -96,7 +96,7 @@ export default class ScratchOrgInfoFetcher {
 
     return  retry(
       async (bail) => {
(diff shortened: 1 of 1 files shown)


azlam-abdulsalam:
We had earlier thought about this,  but this can result in over assignment,  for eg, a developer pool which is provided with 10 scratch orgs,  and are all assigned. Next run of prepare will result in creating 10 additional one's. This can result in starving scratch orgs when you have multiple pools. We ask developers to return used scratch orgs so pools can be replenished

For these scenarios,  we typically create more pools, and pass that as an array to the validate command or provide option to developers to fetch from other pool.

Please see the discussion below

https://github.com/Accenture/sfpowerscripts/discussions/702

While I agree, this can simplify pool management, I am worried on the scale aspect.

 What do you think?

RobinWeymans:
I didn't see this discussion prior to posting. It has some pretty good arguments as to why it's not always a good idea. I'll answer over there 😄

azlam-abdulsalam:
I am closing this, as we will discuss this

PR #737, opened 2021-10-23
Fix #732 Hold a deployment of package till permission set groups are updated

### LINKED ISSUE #732: Packages with PermissionSetGroup fails to deploy
**Describe the bug**

When a package that contains permission set is deployed along with another package that contains PermissionSetGroup, the package with PermissionSetGroup will fail to deploy as permission set group is being updated. 

**To Reproduce**

1. Create a package A with Permset
2. Create a package B with Permset Group that depends on Permset which is part of A
3. Notice Package B fails to deploy

**Expected behavior**

Package B should be successfully installed

**Screenshots**

 - OS:  All
 - Version [e.g. CLI Version  eg: 1.6.6]: 8.5.2
 - Salesforce CLI(sfdx cli) Version: 7.120.0
 - CI Platform:  All

**Additional context**

azlam-abdulsalam:
- Add psg update class
- feat(package): fixes #732 , Hold a deployment of a particular package till all PSG is updated


packages/core/src/permsets/PermissionSetGroupUpdateAwaiter.ts
```diff
@@ -0,0 +1,49 @@
+import { Connection } from "@salesforce/core";
+import SFPLogger, { Logger, LoggerLevel } from "../logger/SFPLogger";
+import QueryHelper from "../queryHelper/QueryHelper";
+import { delay } from "../utils/Delay";
+
+
+
+
+const psGroupQuery = `SELECT Id,MasterLabel,Status FROM PermissionSetGroup WHERE Status != 'Updated'`;
```
dieffrei: What if some permissionsetgroup has the status "Failed"? Failed status doesn't affect deployments, only doesn't a allow assignments.

dieffrei: The only problem I see is if there is a permission set group that failed to be recalculated and is not versioned, it can block deployments.

azlam-abdulsalam: Thanks @dieffrei  oh interesting, is there a failed status to it as well?

azlam-abdulsalam: I have updated the query to only focus on 'Updating'

dieffrei: Sounds good
Indicates permission set group recalculation status. Valid values are:
Updated—The group is current.
Outdated—The group requires recalculation.
Updating—The group is in recalculation mode.
Failed—The group recalculation failed.

PR #743, opened 2021-10-28
Refactor PackageManifest class

aly76:
Change method of instantiation of `PackageManifest` to factory method. This allows
the package.xml to be parsed when `PackageManifest` is instantiated, as
constructors cannot be asynchronous. By enforcing the xml to be parsed at the time of object
creation, the instance methods are guaranteed to have access to the manifest.

Fix bug where 'types' is undefined and causes an error trying to reference a
property on type undefined.

Fixes https://github.com/dxatscale/sfp-cli/issues/47

### DIFF
--- a/packages/core/src/package/SFPPackage.ts
+++ b/packages/core/src/package/SFPPackage.ts
@@ -127,7 +127,7 @@ export default class SFPPackage {
 
     // Requires destructiveChangesPath which is set by the property fetcher
     sfpPackage._workingDirectory = SourcePackageGenerator.generateSourcePackageArtifact(
-      sfpPackage._logger, 
+      sfpPackage._logger,
       sfpPackage._projectDirectory,
       sfpPackage._package_name,
       sfpPackage._packageDescriptor.path,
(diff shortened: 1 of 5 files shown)


packages/core/src/package/SFPPackage.ts
```diff
@@ -142,8 +142,8 @@ export default class SFPPackage {
       packageLogger
     ).exec(true);
 
-    let packageManifest:PackageManifest = new PackageManifest(sfpPackage.mdapiDir);
-    sfpPackage._payload = await packageManifest.getManifest();
+    const packageManifest:PackageManifest = await PackageManifest.create(sfpPackage.mdapiDir);
+    sfpPackage._payload = packageManifest.manifest;
     sfpPackage._triggers = packageManifest.fetchTriggers();
```
azlam-abdulsalam, line 147: Do we need a const here, cant we just short circuit it?

azlam-abdulsalam (approved):
Nice!

packages/core/src/package/SFPPackage.ts
azlam-abdulsalam: Ignore me :)

PR #755, opened 2021-11-08
Enable pools to succeed on deployment errors by default

azlam-abdulsalam:
There have been multiple feedback from the community, that prepare breaks quite often when there is a missing setting or a dependency cycle,resulting in scratch orgs not being created and the pools not available. 
The option is to allow partial scratch orgs, rather than nothing at all. In order to enable this, succeedOnDeploymentErrors will be enabled as the default behaviour ﻿


aly76 (approved):
Looks good

PR #757, opened 2021-11-10
fix(npmtag): deprecate usage of npmtag for artifacts

### LINKED ISSUE #687: Inconsistent artifact versions in releases using NPM tags
**Describe the bug**
Release definitions with artifact versions specified using NPM tags may lead to a release where different artifact versions are installed to each org. This happens when a new artifact is published with the NPM tag, before the release has a chance to complete for every org.

**To Reproduce**

1. Define release definition using NPM tags e.g. 'main'
2. Release to Org 1
3. Build & publish new artifact with the same NPM tag
4. Release to Org 2 

**Expected behavior**
A release should be made up of a set of consistent artifact versions that are released to each org.

**Workaround**
Use `LATEST_TAG` or exact version number in release definitions. The former will only consider git tags that were created before the HEAD commit, which should be consistent in a release, across orgs.

azlam-abdulsalam:
npmTag option will be deprecated across sfpowerscripts to reduce complexity while
handling artifacts. All artifacts will be picked up using LATEST_GIT_TAG that can be reached from a
given branch

BREAKING CHANGE: Pool Schema no longer supports npmTag

fix #687


packages/sfpowerscripts-cli/src/commands/sfpowerscripts/orchestrator/publish.ts
```diff
@@ -321,14 +325,7 @@ export default class Promote extends SfpowerscriptsCommand {
...
     let tag: string;
-    if (this.flags.npmtag) {
-      tag = this.flags.npmtag;
-    } else if (packageMetadata.branch) {
-      tag = packageMetadata.branch
-    } else {
-      throw new Error(`Artifact ${packageName} ${packageVersionNumber} does not contain branch info. Please provide --npmtag flag explicitly, or re-build the artifact with the --branch flag.`);
-    }
-
+    tag = packageMetadata.branch
```
aly76: Should guard against undefined/null tag

aly76 (changes requested):

packages/sfpowerscripts-cli/src/commands/sfpowerscripts/orchestrator/publish.ts
azlam-abdulsalam: is there a chance for branch to be null? as branch is a mandatory flag now

aly76 (dismissed):

aly76:
This PR introduced a breaking change to the CLI. The artifactFetchScript for the `prepare` command receives the artifact version through variable $2, and the path is now given by variable $3

PR #766, opened 2021-11-19
fix(cli-table): set cli-table dependency to version 0.3.6

aly76:
Version 0.3.7 and 0.3.8 are failing to strip ansi


azlam-abdulsalam (approved):
okie

PR #770, opened 2021-11-24
Added no env variable to prevent sfpowerkit header logger appearing

henry88lay:
Added an ENV Variable conditional to prevent the header log from sfpowerkit appearing on sfpowerscripts


azlam-abdulsalam (changes requested):
Closing this as its an incorrect change

PR #775, opened 2021-12-06
Feature/setup-sfpowerkit-noheader-env-var

henry88lay:
Set up an environment variable for sfpowerkit to not display the header and ran tests on commands.


packages/sfpowerscripts-cli/src/SfpowerscriptsCommand.ts
```diff
@@ -2,6 +2,7 @@ import { SfdxCommand } from "@salesforce/command";
 import { OutputFlags } from "@oclif/parser";
 import SFPStatsSender from "@dxatscale/sfpowerscripts.core/lib/stats/SFPStatsSender";
 import * as rimraf from "rimraf";
+import child_process = require("child_process");
```
azlam-abdulsalam: Unused?

henry88lay: yeah its unused i am going to remove it

azlam-abdulsalam:
@henry88lay  can you add a condition on the variable, so that header is enabled if the loglevel is set to debug or below? This could be helpful to understand what is the version of sfpowerkit being used

henry88lay:
> @henry88lay can you add a condition on the variable, so that header is enabled if the loglevel is set to debug or below? This could be helpful to understand what is the version of sfpowerkit being used

good idea i will look into getting that implemented.

PR #777, opened 2021-12-07
Publish aggregated work items and commits metric in release

aly76:
Sum all of the work items and commits across builds that are part of the same
release.


azlam-abdulsalam (approved):
okie

PR #778, opened 2021-12-07
Support for POSIX systems that not not using bash as default shell

### LINKED ISSUE #745: Error when post deployment script in Bitbucket
**Describe the bug**
When our post-deployment script is run in our Bitbucket pipeline we get the following error:

Error: 
/bin/sh: bash: not found

**To Reproduce**
1. Add the following to the sfdx-project.json
`
        {
            "path": "./testdata",
            "package": "TestData",
            "versionNumber": "1.0.0.0",
            "type": "data",
            "postDeploymentScript": "scripts/bitbucket/dataPostDeployment.sh",
            "skipDeployOnOrgs": [
                "qa",
                "staging",
                "prod"
            ]            
        }`

**Expected behavior**
The scripts/bitbucket/dataPostDeployment.sh runs successfully like it does when run locally.

**Desktop (please complete the following information):**

- OS: Windows 10
- Node Version
- Version of sfpowerkit[e.g 1.45.0]: @dxatscale/sfpowerscripts 8.4.0
- Salesforce Org Version [ In case of pre release window ]

aly76:
Replace bash with sh, which points to the default Bourne shell used by the OS.

Closes #745

### DIFF
--- a/packages/core/src/package/packageInstallers/PackageInstallationHelpers.ts
+++ b/packages/core/src/package/packageInstallers/PackageInstallationHelpers.ts
@@ -12,7 +12,7 @@ export default class PackageInstallationHelpers {
   ) {
     let cmd: string;
     if (process.platform !== "win32") {
-      cmd = `bash -e ${script} ${sfdx_package} ${targetOrg}`;
+      cmd = `sh -e ${script} ${sfdx_package} ${targetOrg}`;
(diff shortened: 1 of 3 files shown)


aly76:
This change may break scripts written in bash and where sh does not point to bash.

PR #780, opened 2021-12-07
chore(package-diff)!: scratch definition file no longer considered a change

aly76:
Unlocked packages are no longer considered as changed if the scratch definition
file has changed.


azlam-abdulsalam (dismissed):
okie. Lets have a key call out for this in the release notes for this change

PR #790, opened 2021-12-14
Fix base branch resolution

aly76:
- Fix base branch resolution
- Remove incorrect comment


azlam-abdulsalam (approved):
Thanks looks good

PR #841, opened 2022-02-12
refactor(alias): refactor AliasListImpl to use @salesforce/core as opposed to cli

azlam-abdulsalam:
sfpowerscripts depend on sfdx-cli's directly, This was known as a bad practice at the onset, the lack of visibility into how it was coded in sfdx and considering the time taken to rebuild, it was decided to be  executed over  sfdx-cli, however to better control limited subset required by sfpowerscripts, it is better to refactor to use @salesforce/core and other libraries such as
@salesforce/source-deploy-retrieve provided by salesforce


aly76 (approved):

aly76:
`Aliases` has been deprecated in @salesforce/core, need to monitor for new implementations

PR #847, opened 2022-02-20
refactor(source-convert): use source-deploy-retrieve instead of cli to do convert to mdapi

azlam-abdulsalam:
Remove direct cli to source:convert and use sdr instead


packages/core/tests/package/packageFormatConvertors/SourceToMDAPIConvertor.test.ts
```diff
@@ -0,0 +1,21 @@
+import { jest, expect } from "@jest/globals";
+import SourceToMDAPIConvertor from "../../../src/package/packageFormatConvertors/SourceToMDAPIConvertor";
+import {
+  ComponentSet,
+  MetadataConverter,
+} from "@salesforce/source-deploy-retrieve";
+
+
+describe("Given a source directory, it should be able to convert to mdapi directory", () => {
+  it("should provide the path to a mdapi directory", async () => {
```
aly76: This test does not provide any value. It's asserting that a mocked return value is valid.

aly76 (dismissed):

packages/core/tests/package/packageFormatConvertors/SourceToMDAPIConvertor.test.ts
azlam-abdulsalam: Removed tests

aly76 (approved):

PR #850, opened 2022-02-22
refactor(package-list): refactor to use query instead of wrapping around sfdx force:package:list

azlam-abdulsalam:


packages/core/src/package/packageQuery/PackageFetcher.ts
```diff
@@ -0,0 +1,29 @@
...
+{
+
+  
+  constructor(
+    private conn: Connection
+  ) {}
+
+  public async listAllPackages()
+  {
+    let records = await QueryHelper.query<any>(query, this.conn, true);
```
aly76: Record type should be provided as type parameter, instead of `any`, so that the accessible fields on the `Package2` object are known.

aly76 (dismissed):

aly76 (approved):

vuha-acn (commented):
@azlam-abdulsalam  - For each new salesforce release, do we need to update the API Version (eg. v53.0) each time and regression test since is dynamic? Can we have global variable to control this to update in 1 single place instead?

azlam-abdulsalam:
@vuha-acn We don't use api version (unless in tests), it is set to whatever the target api version if the org is

PR #851, opened 2022-02-22
fix: misleading error message for prepare cmd

aly76:
Fix misleading error that shows missing maxRevision.json for deployment failures


azlam-abdulsalam (approved):
LGTM

PR #865, opened 2022-02-28
Remove backward compatibility layer from artifacts

azlam-abdulsalam:
- refactor(artifact-file-type): remove backward compatibility code, as everything is a zip or npm
 - refactor(artifact-migrator): remove need for artifact migrator, as now everything has migrated to v2 
 - refactor(artifact-installation): refactor Artifact Record keeping into something owned by a Org


packages/core/src/org/SFPOrg.ts
```diff
@@ -0,0 +1,217 @@
...
+                return records;
+            } catch (error) {
+                throw new Error(
+                    'Unable to fetch any sfpowerscripts artifacts in the org\n' +
+                        '1. sfpowerscripts package is notinstalled in the org\n' +
+                        '2. The required prerequisite object is not deployed to this org\n'
+                );
+            }
+        } else {
+            return this.installedArtifacts;
```
aly76: I'm not sure that we should cache the state of the org, unless we intend each instance of `Org` to be the state of the org at a particular moment in time. I think that we should be able to query the current state of the org, using the same instance.

aly76: if the consumer does not want to query again, they should use the return value from the first call.

packages/core/src/artifacts/ArtifactFilePathFetcher.ts
```diff
@@ -26,23 +26,16 @@ export default class ArtifactFilePathFetcher {
 
         let artifacts: string[] = this.findArtifacts(artifactDirectory, sfdx_package);
 
-        if (artifacts.length === 0) {
```
aly76, line 29: This merge request will cause a breaking change

aly76 (changes requested):

packages/core/src/org/SFPOrg.ts
```diff
@@ -0,0 +1,217 @@
...
+    public async listAllPackages() {
+        if (this.isDevHubOrg()) {
+            let records = await QueryHelper.query<PackageTypeInfo>(packageQuery, this.getConnection(), true);
+            records.forEach((record) => {
+                record.IsOrgDependent =
+                    record.ContainerOptions === 'Managed' ? 'N/A' : record.IsOrgDependent === true ? 'Yes' : 'No';
+            });
+
+            return records;
+        } else throw new Error('Only applicable in a DevHub');
```
aly76: Can the error message be more descriptive as to what is only applicable in devhub?

packages/core/src/artifacts/ArtifactMigrator.ts
```diff
@@ -1,157 +0,0 @@
-import child_process = require('child_process');
```
aly76, line 1: This is a breaking change, will mean that `SfpowerscriptsArtifact__c` will no longer be migrated to `SfpowerscriptsArtifact2__c`

packages/core/tests/org/ArtifactsToOrg.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    });
+
+    it('Update a sfpowerscripts artifact and resulting an error,should throw an exception', async () => {
+        let org = await createOrg();
+
+        let records: AnyJson = {
+            records: [
+                {
+                    attributes: {
+                        type: 'Sfpowerscriptspackage__c',
```
aly76: SfpowerscriptsArtifact2__c

packages/core/tests/org/ArtifactsToOrg.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    });
+
+    it('Update a sfpowerscripts artifact, installing a newer version of it', async () => {
+        let org = await createOrg();
+
+        let records: AnyJson = {
+            records: [
+                {
+                    attributes: {
+                        type: 'Sfpowerscriptspackage__c',
```
aly76: SfpowerscriptsArtifact2__c

aly76 (changes requested):

packages/core/src/org/SFPOrg.ts
azlam-abdulsalam:
```suggestion
        } else throw new Error('Package Type Information can only be fetched from a DevHub');
```

packages/core/tests/org/ArtifactsToOrg.test.ts
azlam-abdulsalam:
```suggestion
                        type: 'SfpowerscriptsArtifact2__c',
```

packages/core/tests/org/ArtifactsToOrg.test.ts
azlam-abdulsalam:
```suggestion
                        type: 'SfpowerscriptsArtifact2__c',
```

azlam-abdulsalam (commented):
Applied changes

aly76 (approved):

PR #866, opened 2022-02-28
docs(decision record): implementation of source tracking on prepare command

aly76:


azlam-abdulsalam:
@aly76 Could we also factor in the size of the static resource? As earlier it was a single file, now this would be the git tree

aly76:
> @aly76 Could we also factor in the size of the static resource? As earlier it was a single file, now this would be the git tree

This is an interesting point and potential blocker, as large git repositories can be hundreds of MB and the a static resource has a maximum size of 5MB.

PR #881, opened 2022-03-05
Activate Synchronous compile on during Prepare

azlam-abdulsalam:
#### Description

Scratch Orgs prepared from pool are typically built without 'Synchronous Compile on Deploy'. This makes deployment faster especially reducing the need to compile on every packages. However, this has side effect when pools are used during validate, where an async apex test may trigger compilation of test classes, which result in inaccurate code coverage. In order to counter this, it is better to have changes validated during validate perform compilation immediately. This feature enables
synchronous compile as default. Please note, this feature is only activated at the end of prepare
job, so that it doesn't slow prepare itself (as prepare doesn't need tests)

#### Checklist

- [ ] Updates to documentation
- [x] I have tested  that changes works
- [x] New and existing unit tests pass locally with my changes


azlam-abdulsalam:
Closing this PR, will raise after DR is approved

PR #901, opened 2022-03-22
Fix regex for validing auth URL

aly76:
Add `=` char to refresh token


azlam-abdulsalam:
Closing this as this will be addressed in #902

PR #911, opened 2022-03-25
Add diffcheck flag to validate commands

### LINKED ISSUE #800: Add support for diffcheck parameter in validateAgainstOrg command
A recent update (#759) disabled the diff check in the validateAgainstOrg command.  For those organizations that have yet to migrate to scratch orgs (for all the various reasons that exist), it would be helpful to still use the validateAgainstOrg command and have a `--diffcheck` parameter similar to build and quickBuild.  The command can still default this to false, but for organizations that depend on sandboxes for validation, it helps to skip untouched packages to speed up the validation process.

azlam-abdulsalam:
#### Description

Based on the contribution by @jhawk-4 in #801

- Add the ability to disable  diffcheck  that is performed during validation by using a flag on the $sfdx sfpowerscripts:orchestrator:validate and $sfdx sfpowerscripts:orchestrator:validateAgainstOrg commands.

- Added a flag to disable committing artifact records in the org. This will prevent the issue when a temporary merge (as in case of PR) is being used to validate on a reused org.

#### Checklist

- [X] Tested changes?
- [X] Unit Tests new and existing passing locally?


packages/core/src/package/packageInstallers/InstallPackage.ts
```diff
@@ -139,15 +140,22 @@ export abstract class InstallPackage {
         });
     }
 
+    //Set this to disable whethere info about the artifact has to be recorded in the org
+    public set disableArtifactCommit(disableArtifactCommit: boolean) {
```
aly76: Something wrong with this function semantically or logically. As a reader, I would expect `disableArtifactCommit` to only set `isArtifactToBeCommitted` to false, but it's confusing as it also accepts a param which either causes it to go ahead with the disable operation or enable the functionality - which is a toggle semantically.

aly76: Incorrect usage of setter, where the setter is not named after the prop that it sets.

aly76 (changes requested):

aly76 (approved):

PR #930, opened 2022-04-07
fix(source-tracking): clean up mpd to prevent source tracking lib to gather status of all files

azlam-abdulsalam:
#### Description

Possibly source tracking lib use statusMatrix to gain information about all files when the
repository is initiated. This is gathered from the projects path at the point when the library is
intiatlized. As sfpowerscripts fixes source tracking one after the another, this is not necessary
and is resulting in some errors (statusMatrix unable to find files.. though no functional impact)
and also unnecessary computation. This fix is to clean up the sfdx-project.json to only carry the
package that is currently being fixed up for source tracking

#### Checklist

- [x] Tested changes?

### DIFF
--- a/packages/core/src/scratchorg/pool/ClientSourceTracking.ts
+++ b/packages/core/src/scratchorg/pool/ClientSourceTracking.ts
@@ -89,12 +89,10 @@ export default class ClientSourceTracking {
                     LoggerLevel.INFO,
                     this.logger
                 );
-                SFPLogger.log(`Analyzing package ${COLOR_KEY_MESSAGE(artifact.Name)}`, LoggerLevel.INFO, this.logger);
-
-                //Reset the project json before going for full track
-                await git.raw(['restore', 'sfdx-project.json']);
+                SFPLogger.log(`Analyzing package ${COLOR_KEY_MESSAGE(artifact.Name)}`, LoggerLevel.INFO, this.logger);         
                 // Checkout version of source code from which artifact was created
-                await git.checkout(artifact.CommitId__c);
+                await git.checkout(['-f',artifact.CommitId__c]);
+               
                 SFPLogger.log(
                     `Version pushed while preparing this org is ${artifact.Version__c} with SHA ${artifact.CommitId__c}`,
                     LoggerLevel.INFO,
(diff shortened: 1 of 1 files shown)


packages/core/src/scratchorg/pool/ClientSourceTracking.ts
```diff
@@ -60,31 +60,49 @@ export default class ClientSourceTracking {
...
+                SFPLogger.log(EOL, LoggerLevel.INFO, this.logger);
+                SFPLogger.log(
+                    COLOR_HEADER(`Package ${count} of ${sfpowerscriptsArtifacts.length}`),
+                    LoggerLevel.INFO,
+                    this.logger
+                );
+                SFPLogger.log(`Analyzing package ${COLOR_KEY_MESSAGE(artifact.Name)}`, LoggerLevel.INFO, this.logger);
+
+                //Reset the project json before going for full track
+                await git.raw(['restore', 'sfdx-project.json']);
```
aly76: Cleaner if we just force checkout

packages/core/src/scratchorg/pool/ClientSourceTracking.ts
```diff
@@ -60,31 +60,49 @@ export default class ClientSourceTracking {
...
+            this.cleanupSFDXProjectJsonTonOnePackage(tempDir.name, sfpowerscriptsArtifacts[0].Name);
+
             const project = await SfdxProject.resolve(tempDir.name);
 
             // Create local source tracking files in temp repo
             const tracking = await SourceTracking.create({
                 org: this.org,
                 project: project,
             });
-            tracking.ensureLocalTracking();
```
aly76, line 71: Why is the initialization for source tracking removed?

azlam-abdulsalam: Not required, as the updateLocalTracking calls it anyway

packages/core/src/scratchorg/pool/ClientSourceTracking.ts
azlam-abdulsalam: I looked at the faster option ( restore vs checkout ), as we only change one file

aly76: It's probably faster to do a force checkout than two operations: restore and checkout

azlam-abdulsalam: makes sense, fixing it

PR #1071, opened 2022-06-28
feat(filter): add support for skipping entitlements

### LINKED ISSUE #1070: Add support for filtering entitlements
Entitlements(https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_entitlementprocess.htm) are not CI/CD friendly. It has an  file based versioning which doesnt work with sfpowerscripts 'source' packages, as the entirety of source package gets installed every time rather than what is being changed. As many orgs have entitlements that is ranging from 30-50, this soon becomes too cumbersome in terms of manual steps

sfpowerscripts can help by providing filters, where during a deployment, the org can be introspected with the versionNumber of the entilement being used and proceed to install only if the version number is incremented. This feature can also be turned of globally by plugins.sfpowerscripts.disableEntitlementFilter.

This also demonstrates how we could build filters and the component set be transformed before being deployed to the target org.

azlam-abdulsalam:
Fixes #1070
- Add support for filters before deploying source packages
- Create component set to deploy

#### Checklist

- [x] Tested changes?

### DIFF
--- a/packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
+++ b/packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
@@ -48,15 +48,13 @@ export default class EntitlementVersionFilter implements DeploymentFilter {
                 if (sourceComponent.type.name === registry.types.entitlementprocess.name) {
                     let slaProcessLocal = sourceComponent.parseXmlSync();
 
-                    let slaProcessMatchedByName: SlaProcess = slaProcessesInOrg.find((element: SlaProcess) => {
-                        if (entitlementSettings.enableEntitlementVersioning)
-                            return element.NameNorm == sourceComponent.name;
-                        else return element.Name == sourceComponent.name;
-                    });
-                    if (!slaProcessMatchedByName) {
-                        //Doesnt exist, deploy
-                        modifiedComponentSet.add(sourceComponent);
-                    } else if (
+                    let slaProcessMatchedByName: SlaProcess = slaProcessesInOrg.find(
+                        (element: SlaProcess) => element.Name == slaProcessLocal['EntitlementProcess']['name']
+                    );
+
+                    if (
+                        slaProcessMatchedByName &&
+                        entitlementSettings.enableEntitlementVersioning &&
                         slaProcessLocal['EntitlementProcess']['versionNumber'] > slaProcessMatchedByName.VersionNumber
                     ) {
                         //This is a deployment candidate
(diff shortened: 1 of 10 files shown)


packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
```diff
@@ -0,0 +1,78 @@
...
+                    modifiedComponentSet.add(sourceComponent);
+                }
+                else if (slaProcessLocal['EntitlementProcess']['versionNumber'] > slaProcessMatchedByName.VersionNumber) {
+                    //This is a deployment candidate
+                    //Modify versionMaster tag to match in the org
+                    slaProcessLocal['EntitlementProcess']['versionMaster'] = slaProcessMatchedByName.VersionMaster;
+                    let builder = new XMLBuilder({ format: true, ignoreAttributes: false, attributeNamePrefix: '@_' });
+                    let xmlContent = builder.build(slaProcessLocal);
+                    fs.writeFileSync(sourceComponent.xml, xmlContent);
+                    modifiedComponentSet.add(sourceComponent);
```
zhebinliu: I think we need to check if the version in the XML is the same as the one in the file name(NameNorm), in case the dev just clones the file and only updates the version in the XML.

packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
```diff
@@ -0,0 +1,78 @@
...
+export default class EntitlementVersionFilter implements DeploymentFilter {
+    public async apply(org: SFPOrg, componentSet: ComponentSet,logger:Logger): Promise<ComponentSet> {
+        //Only do if entitlment exits in the package
+        SFPLogger.log(`Filtering Entitlement Process....`,LoggerLevel.INFO,logger);
+        let sourceComponents = componentSet.getSourceComponents().toArray();
+        let isEntitlementFound: boolean = false;
+        for (const sourceComponent of sourceComponents) {
+            if (sourceComponent.type.name === registry.types.entitlementprocess.name) {
+                isEntitlementFound = true;
+                break;
```
zhebinliu, line 22: should we get a list of all entitlement processes here, in case the component set is very large

packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
azlam-abdulsalam: I am not sure, what about the entitlements with no version info or doesnt have versioning enabled? Let me add some tests for that scenario

packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
azlam-abdulsalam: I dont think there is anything we can do..  even the normal one would be a filter, Let me check

packages/core/src/package/componentFilter/EntitlementVersionFilter.ts
azlam-abdulsalam: I have addressed this in the latest commit

zhebinliu (dismissed):

PR #1128, opened 2022-09-27
Issue #978 specify artifacts to be fetched in pool config file

### LINKED ISSUE #978: [New Feature]: Specify Artifacts to be Installed in Pools
**Describe the feature**

- A pool should be able to be created with the desired artifacts. 
- That includes managed package dependencies as well as unlocked, source, or data packages.
- As long as the specified set of artifacts is compiled and able to be built with no dependency issues, any given combination of artifacts should be deployable in pools.

**Assumptions;**
- Project is a mono repo.
- There are no multiple branches maintained per team, feature, domain, etc. The main trunk is the source of truth.

**Why?**
- There are different teams working on the monolith based on business domains. Oftentimes, one team has nothing to do with some of the managed package dependencies as well as unlocked or source packages. 
- Business wants a demo showcasing a specific application domain. We should be able to define a demo pool that has only what needs to be there.

How?
- A new attribute can be exposed as part of pool configuration json where the desired dependencies and artifacts can be provided.
- This can be also done over multiple sfdx-project.json definitions. 
- How to avoid project json conflicts between different teams is a catch here.

domrycz:
#### Checklist

- [x] Tested changes?


azlam-abdulsalam:
Thanks @domrycz  for starting to work on this. As a eventual solution, we would need to add support for tags to all packages ( I will write up a DR shortly) and then the pool config you can mention tags, and we go on installing artifacts that form part of the tag.

As a temporary solution, the above can work. but I believe from a schema perspective, it would be better to have it on the top level rather than on the npm object.  We can filter it after we have downloaded

@zhebinliu  your thoughts

domrycz:
Thanks for checking this PR @azlam-abdulsalam, I will wait for DR then

zhebinliu:
I think as for now, we can utilise the release definition file by adding an extra attribute in the pool config file to specify the release definition file to be used for the pool, if the version is missing for specific packages, we can bring the latest version as we are doing at the moment. and in the long term, the tag will the one to use

azlam-abdulsalam:
@zhebinliu Thats an awesome idea, it would be easy to implement as well, as the code is already available

azlam-abdulsalam:
@domrycz can you proceed ahead with releasedefinion as an optional entity in pool config?

domrycz:
@azlam-abdulsalam I think I can at least try :) I will try to do something in this week.

domrycz:
@azlam-abdulsalam @zhebinliu can you take a look guys?

zhebinliu:
> @azlam-abdulsalam @zhebinliu can you take a look guys?

Hi, thanks for your commits. The PR failed to pass the test, can you please do a few more local tests to ensure the code is working?

domrycz:
@zhebinliu PR fixed

packages/sfpowerscripts-cli/src/impl/release/ReleaseDefinition.ts
```diff
@@ -45,6 +45,11 @@ export default class ReleaseDefinition {
         return releaseDefinition;
     }
 
+    public static async getArtifactsFromReleaseDefinitionFile(pathToReleaseDefinition: string) {
```
azlam-abdulsalam: Do we really need this helper? Cant we just consume it from the release definition just returned.

I think its confusing, as its called artifacts and on quick glance, i thought it was a real artifacts :)

azlam-abdulsalam (commented):
Added a question regarding ReleaseDefintion

azlam-abdulsalam:
Guys what do you think about using the release definition generator's config file as the input, this will allow a lot of flexiblity,
https://docs.dxatscale.io/sfpowerscripts/release/release-definition-generator
We use this defintion to generate a release defintion on the fly

domrycz:
@azlam-abdulsalam I've added new changes to the PR. I'm using release def config file to specify artifacts - it seems it's indeed more clear solution. Please check :)

domrycz:
@azlam-abdulsalam @zhebinliu hey guys, how is it going with this? Is there a chance that this will be included in the next release?

azlam-abdulsalam:
@domrycz  yes, working on this one, this will be tied to multiple validate modes

azlam-abdulsalam:
I have merged this into branch for #1133, Raising a PR shortly

PR #1134, opened 2022-10-03
Field history tracking resolution DR

gnemiq:
#### Checklist

- [x] Tested changes?


rody:
Adding here the summary of the discussion with @azlam-abdulsalam  on Slack:

The discussion was about not introducing another file format to maintain but use the metadata files to manage the history tracking. This to ensure that we don't add another "way of working" from a dev perspective.

>  1. use the "normal" metadata (i.e. CustomField metadata) as the source of truth (no changes from a dev perspective)
>  2. exclude the standard and prefixed metadata (i.e. metadata that we don't own) from the package/deployment
>  3. from all the metadata (owned and not-owned) generate an input (CSV,JSON,....) for the "alignment" script.
>  4. after package installation/deployement run the script on-platform to update only the fields that are ignored during installation (ex. the tracking history) and the modifiable fields on non-owned metadata (with the input from 3).
>   
>This will prevent the dev from having another file to maintain. They can still use the normal flow (update on SO, retrieve, commit).

azlam-abdulsalam (approved):
okie

PR #1154, opened 2022-11-17
Fixes #978 and #1133  Multiple changes to Prepare/Validate

### LINKED ISSUE #978: [New Feature]: Specify Artifacts to be Installed in Pools

### LINKED ISSUE #1133: Validate a package and its dependencies only
sfpowerscripts maybe utilized in scenarios where the packages are relatively independent but served from the same repository such as tech frameworks.  To support this mode, it would be beneficial to have an additional mode during validate which only validate a package along with installing its dependencies (not packages that depend on it)

azlam-abdulsalam:
- Fixes #978  Add support for pools to have selective packages by release config
- Fixes #1133  Multiple validate mode, add support for multiple validate modes  (--mode=(individual|fastfeedback|thorough|ff-release-config|thorough-release-config)

#### Checklist

- [ ] Tested changes?


packages/sfpowerscripts-cli/src/impl/release/ReleaseDefinitionGenerator.ts
```diff
@@ -155,7 +162,10 @@ export default class ReleaseDefinitionGenerator {
         return { artifacts, packageDependencies };
     }
 
-    private async generateReleaseDefintion(artifacts: any, packageDependencies: any, git: Git) {
+    private async generateReleaseDefintion(artifacts: any, packageDependencies: any, git: Git): Promise<ReleaseDefinitionSchema | {
```
domrycz, line 165: typo in definition (defintion) here in the method name and also in variable names in e.g. 96, 107, 166

PR #1156, opened 2022-11-20
Draft: Validation retry support

dieffrei:
#### Checklist

- [ ] Tested changes?


packages/core/src/scratchorg/pool/PoolFetchImpl.ts
```diff
@@ -131,6 +135,7 @@ export default class PoolFetchImpl extends PoolBaseImpl {
                 let allocateSO = await new ScratchOrgInfoAssigner(this.hubOrg).setScratchOrgInfo({
                     Id: element.Id,
                     Allocation_status__c: 'Allocate',
+                    ValidationId__c: this.validationId
```
azlam-abdulsalam, line 138: Looks like this will break in orgs where this field is yet to be deployed, we need to handle this

azlam-abdulsalam (changes requested):
@dieffrei  Can you add a DR along with the PR and also see my query

packages/core/src/scratchorg/pool/PoolFetchImpl.ts
vuha-acn: Do we need to upgrade the **sfpower-scratchorg-pool** to a new version 2.1 to reflect this new `ValidationId__c` field or are we adding error handling to skip depending on this field?  We can draft standard manual updates as well to that field in the repo for the **sfpower-scratchorg-pool**

azlam-abdulsalam: @vuha-acn  thats correct, we need an update to 2.1, and also need to handle others on 2.0

dieffrei: Thank you, I will check that

azlam-abdulsalam:
Currently closing this as it has become stale!

PR #1179, opened 2022-12-19
feat(fht): enable field history tracking support

azlam-abdulsalam:
This issue adds ability to generate a set of components that need to be post deployed for enabling
field history tracking for unlocked packages or for specifically to remove deuplication of managed
package comopoonents.

#### Checklist

- [x] Tested changes?


packages/core/src/package/analyser/FHTAnalyzer.ts
```diff
@@ -0,0 +1,73 @@
...
+        );
+
+        //read components mentioned in yaml
+        if (fs.existsSync(fhtYamlPath)) {
+            //convert yaml to json
+            fhtFields = yaml.load(fs.readFileSync(fhtYamlPath, { encoding: 'utf-8' }));
+        }
+
+        //Load component Set
+        let componentSet = ComponentSet.fromSource(
```
MengSQi, line 29: if we pass the componentSet from the builder, we only need to generate it once and all other analysers/other possible functions in the builder can use it

packages/core/src/package/postDeployers/FHTEnabler.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    public async gatherPostDeploymentComponents(
+        sfpPackage: SfpPackage,
+        conn: Connection,
+        logger: Logger
+    ): Promise<ComponentSet> {
+        
+
+        //Generate component sets
+        let componentSet = ComponentSet.fromSource(path.join(sfpPackage.workingDirectory, sfpPackage.packageDirectory));
```
MengSQi, line 32: consider passing it from the installer

packages/core/src/package/analyser/FHTAnalyzer.ts
gnemiq: if we pass the componentSet from the builder, we only need to generate it once and all other analysers/other possible functions in the builder can use it

packages/core/src/package/postDeployers/FHTEnabler.ts
gnemiq: consider passing it from the installer

packages/core/src/package/analyser/FHTAnalyzer.ts
```diff
@@ -0,0 +1,73 @@
...
+            path.join(sfpPackage.workingDirectory, sfpPackage.projectDirectory, sfpPackage.packageDirectory)
```
gnemiq: check path exists

azlam-abdulsalam: not required as these are guaranteed from SfpPackage, it will be set to blank if not set

packages/core/src/package/analyser/FHTAnalyzer.ts
azlam-abdulsalam: Thats correct, will have a look

packages/core/src/package/analyser/FHTAnalyzer.ts
gnemiq: test failing from this part 
image

packages/core/src/package/postDeployers/FHTEnabler.ts
azlam-abdulsalam: Fixed

packages/core/src/package/postDeployers/FHTEnabler.ts
```diff
@@ -0,0 +1,99 @@
...
+            let customFieldFetcher: CustomFieldFetcher = new CustomFieldFetcher(logger);
+            let sfpOrg = await SFPOrg.create({ connection: conn });
+            let fetchedCustomFields = await customFieldFetcher.getCustomFields(sfpOrg, fieldList);
+
+            //Modify the component set
+            for (const sourceComponent of fetchedCustomFields.components.getSourceComponents()) {
+                let sourceComponentXml = await sourceComponent.parseXml();
+
+                if (sourceComponent.type.name == registry.types.customobject.children.types.customfield.name) {
+                    sourceComponentXml['CustomField']['trackHistory'] = true;
```
gnemiq, line 80: need to enable history tracking for the object as well (sourceObjectXml['CustomObject']['enableHistory']) in the case the object is not fht enabled in the org

azlam-abdulsalam: Will do that on a subsequent Pr, as this one exclusively looks at custom fields. Pulling custom object definition can be bit tricky

zhebinliu (approved):

PR #1195, opened 2023-01-16
fix: 🐛 publish only sfp related package tags

### LINKED ISSUE #1183: Make :publish command to push only packages version tags
**Describe the bug**
In our build process, we use floating tags to specify the latest successful build commit. In some cases, we merge multiple release branches at a time. As a result, multiple publish jobs are executed, and they push all tags from the repo (if I understand it correctly). As we have floating tags, it might happen that one push tried to push a floating one into the old position (commit), resulting in job failure. 

**To Reproduce**
There is no standard way to reproduce it but you would emulate it.
- Create a tag on some old commit of the trunk. "lastest-successful-build-main" for example
- Trigger a build process
- Wait till the build job started
- Quickly re-create the tag with the same name on the last commit and push it wish force param to override the position
- Publish commit should fail (see screenshot)

**Expected behavior**
- Publish command should not publish tags which are not relevant to packages' versions

**Screenshots**
Screenshot 2022-12-19 at 14 00 15

 - OS: not relevant
 - Version not relevant
 - Salesforce CLI(sfdx cli) Version: not relevant
 - CI Platform: GitLab

zhebinliu:
pubish command will only pick the tags that is generated by the artifacts and publish. also add support to create tag based on the artifacts commit id.

✅ Closes: 1183

#### Checklist

- [x] Tested changes?

### DIFF
--- a/packages/core/src/git/Git.ts
+++ b/packages/core/src/git/Git.ts
@@ -87,14 +87,31 @@ export default class Git {
         }
     }
 
-    async pushTags() {
-        await this._git.pushTags();
+    async pushTags(
+        tags?:{ 
+            name: string 
+        } [] 
+        ) {
+        if(!tags){
+            await this._git.pushTags();
+        }else{
+            for(let tag of tags){
+                await this._git.push('origin',tag.name)
+            }
+            
+        }
     }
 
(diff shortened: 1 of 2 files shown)


azlam-abdulsalam (approved):
lgtm

PR #1263, opened 2023-03-09
fix(deploy): ensure max retry count is set to 10

### LINKED ISSUE #1262: maxretrycount flag not working for package deployment
**Description:**
When deploying packages, the maxretrycount flag (in the prepare command) does not appear to be functioning properly. Despite setting the maxRetryCount value to 4 in the configuration files, the command only retries twice before failing in some cases.

**Expected behavior:**
When a package deployment fails, the command should automatically retry the specified number of times as set by the maxRetryCount value. In this case, it should stop retrying after 4 attempts to install the package.

**Platform Details:**
 - window 10
 - sfpowerscipt version: 20.25.9
 - Salesforce CLI Version: 7.186.2

azlam-abdulsalam:
deploy doesnt respect max retry count, it is always set at 2 retries, as it was incorrectly set. The
fix sets a  max default of 10 with a break if earlier if the maxRetry is lower

fixes #1262

#### Checklist

- [ ] Tested changes?

### DIFF
--- a/packages/sfpowerscripts-cli/src/impl/deploy/DeployImpl.ts
+++ b/packages/sfpowerscripts-cli/src/impl/deploy/DeployImpl.ts
@@ -241,7 +241,7 @@ export default class DeployImpl {
                             } else return false;
                         }
                     },
-                    { retries: 1, minTimeout: 2000 }
+                    { retries: 10, minTimeout: 2000 }
                 );
 
                 if (packageInstallationResult.result === PackageInstallationStatus.Succeeded) {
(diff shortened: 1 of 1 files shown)


packages/sfpowerscripts-cli/src/impl/deploy/DeployImpl.ts
```diff
@@ -241,7 +241,7 @@ export default class DeployImpl {
                             } else return false;
                         }
                     },
-                    { retries: 1, minTimeout: 2000 }
+                    { retries: 10, minTimeout: 2000 }
```
azlam-abdulsalam:
```suggestion
                    { retries: 10, minTimeout: 30000 }
```

PR #1306, opened 2023-05-12
fix(apexlink): direct conversion to json breaks on large packages

azlam-abdulsalam:
This fix is to ensure apex link's output is not getting truncated when directly fetched as a result
of a process execution. It now writes to a file and then the file is converted to JSON

#### Checklist

- [x] Tested changes?


vuha-acn (approved):
Ok

PR #1418, opened 2023-10-09
docs(dr): propose changes to aliasfy package

azlam-abdulsalam:
Add a DR proposing change to aliasfy package to eliminate redundancy

### Summary generated by Reviewpad on 09 Oct 23 02:47 UTC
This pull request proposes changes to the aliafy package in order to eliminate redundancy. The current implementation of aliasified packages introduces redundancy by duplicating the contents of the 'default' directory across multiple environment-specific directories. The proposed solution is to introduce a layer of inheritance for aliasified packages, where the contents in the 'default' directory will be merged with the contents in the `<alias>` directory if an alias is matched. By implementing this inheritance mechanism, the PR aims to reduce redundancy, simplify maintenance, and minimize the scope for errors.

#### Checklist

- [ ] Tested changes?

### DIFF
--- a/decision records/deployments/003-aliasfy-eliminate-redundancy.md
+++ b/decision records/deployments/003-aliasfy-eliminate-redundancy.md
@@ -1,0 +1,55 @@
+# Eliminating Redundancy in Aliasified Source and Data Packages
+
+* Status: Proposed
+* Deciders: Azlam, Vu, Zhebin
+* Date: 
+
+## Context and Problem Statement
+
+The current implementation of aliasified packages in sfpowerscripts supports a 'default' directory and environment-specific directories (e.g., `dev`, `sit`, `st`). While this structure ensures flexibility, it introduces redundancy. The contents of the 'default' directory often have to be duplicated across multiple environment-specific directories. This leads to increased maintenance complexity and greater risk of errors.
+
+Existing Structure:
+
+```
+src-env-specific-alias-post
+└── main
+    ├── default
+    |   └── <contents>   
+    ├── dev
+    │   └── <contents>   
+    ├── sit
+    │   └── <contents>   
+    └── st
+        └── <contents>   
+```
+
+## Decision 
+
+To alleviate the redundancy and error-prone nature of the current approach, the proposal is to introduce a layer of inheritance for aliasified packages.
+
+Revised Structure:
+
+```
+src-env-specific-alias-post
+└── main
+    ├── default
+    |   └── <base_contents>   
+    ├── dev
+    │   └── <override_contents>   
+    ├── sit
+    │   └── <override_contents>   
+    └── st
+        └── <override_contents>   
+```
+
+- sfpowerscripts will continue to try to match the `alias` as it does today.
+- If an alias matches, the contents in the `<alias>` directory will be merged with the contents in the `default` directory.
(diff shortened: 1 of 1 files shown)


rody (approved):

decision records/deployments/003-aliasfy-eliminate-redundancy.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+- sfpowerscripts will continue to try to match the `alias` as it does today.
+- If an alias matches, the contents in the `<alias>` directory will be merged with the contents in the `default` directory.
+  - If there's a conflict, the `<alias>` directory takes precedence.
+- If the alias isn't found, it will fall back to the `default` directory.
+- If neither is found, an error will be thrown.
+
+
+- To ensure the changes do not disrupt exising users of the aliasfy feature, this enhanced aliasfy package feature has to be explictly enabled as 'aliasfyv2'in the sfdx-project.json
```
rody, line 52: Should we do it the other way around? as in you need to specify V1 if you want to remain on the current thing, so that in the future we can just deprecate V1 without having to maintain this flag?

azlam-abdulsalam: I like that idea, but that means if the users don't see the announcement, they might inadvertently deploy 'default' components into production.

ruslan-kurchenko: I agree with @azlam-abdulsalam . We have to be backward compatible. I bet most of the org user `:latest` image (like we at K).

I would suggest being more specific on the name of this enhancement as it provides certain features, not many - inheritance.

Options:
- `alisifiedInheritence`
- `aliasfyInheritence`
- `enableAlisifiedInheritence`
- `enableAlisifyInheritence`

ruslan-kurchenko:
Great enhancement. We recently talked about it with the team as we started to accumulate so many env variables that we need to maintain.

ruslan-kurchenko (approved):

dieffrei-psignite:
Nice!!

pogilvieCB:
This a a great idea!

decision records/deployments/003-aliasfy-eliminate-redundancy.md
azlam-abdulsalam: or may be better, we treat it as a default behaviour for all package? @ruslan-kurchenko  @rody , if it sees an aliasified folder, it will start doing this behaviour across any source package

crazynammer: @azlam-abdulsalam when you refer to all packages, do you mean that some repos would have like a core-crm source package folder and within that have specific dev, sit, uat alias packages inside that package?

crazynammer (commented):
I like the idea. We'll just have to keep track of all the toggles for aliasfyv2 in the documentation and any other items like this in the future roadmap.

decision records/deployments/003-aliasfy-eliminate-redundancy.md
azlam-abdulsalam: @crazynammer  that's correct, any packages can have env specific overrides, but at the same time, this idea of aliasified envs is conflicting with our notion of detaching tests from dedicated environments,

PR #1421, opened 2023-10-11
feat: use node 18 LTS instead of the latest 20.x version

rody:
### Summary generated by Reviewpad on 11 Oct 23 01:54 UTC
This pull request includes two patches:

1. The first patch updates the Dockerfile for the sfpowerscripts image to use Node.js 18 LTS instead of version 20.x.

2. The second patch modifies the Dockerfile to remove the yarn/sf cache in order to reduce the image size.

Overall, these changes aim to improve the Docker image for the sfpowerscripts build system in the Salesforce environment.

#### Checklist

- [X] Updates to Decision Records considered?
- [X] Updates to documentation at [DX@Scale Guide](https://github.com/dxatscale/dxatscale-guide) considered?
- [X] Tested changes?

# Rationale

We are not dependent on bleeding edge features from nodejs and a few occasional errors have been reported with the image using v20.x. Therefore, it makes sense to use the LTS version of nodejs for our image.

# Image size reduction

@ethan-sargent analyzed the behaviour of `sf plugins install` and discovered that sf was using a specific cache for yarn. The second commit empties this cache after installing the extra plugins as it is not needed. This change saves ~1.4Gb in the image.


azlam-abdulsalam (approved):
okie

PR #1422, opened 2023-10-13
fix(profile): add ability to reset cache when the command is executed against an org multiple times

### LINKED ISSUE #1413: Reset the Sfpowerkit on a different target org or include the org name into the cache
**Describe the bug**
Currently the profile reconcile command does not clean all metadata existing to the connected org. This happens only if you execute the command twice and switch the org between the first run and second run.

**To Reproduce**
1. Connect to Org A
2. Create a new field (Account.Test__c)
3. Add this field to the Admin Profile
4. Run the `sfp profile reconcile` for the Admin-Profile
5. Switch to Org B (without the Test-Field)
6. Run the `sfp profile reconcile` with target Org B
7. Field was not cleared, because the Sfpowerkit has still the Account.Test__c-Field from Org A in the cache

**Expected behavior**
The Sfpowerkit-Cache resets after each change of target org or adds the org alias to the cache.

**Screenshots**

 - OS: MacBook M2
 - Version: @dxatscale/sfpowerscripts/24.5.0 darwin-arm64 node-v20.6.1
 - Salesforce CLI(sfdx cli) Version: @salesforce/cli/2.9.8 darwin-arm64 node-v20.6.1
 - CI Platform: GitHub and Local

**Additional context**
If I add the cache reset to the command, the reconcile works correctly or delete the current cache table:
```
//sfpowerscripts/packages/sfpowerscripts-cli/src/commands/profile/reconcile.ts:86
Sfpowerkit.resetCache();
```

azlam-abdulsalam:
The PR adds an additional flag --resetcache to merge/reconcile which provides ability to reset cache
as required among subsequent runs

fixes #1413

### Summary generated by Reviewpad on 17 Oct 23 01:45 UTC
This pull request includes two patches.

The first patch fixes an issue in the profile merge and reconcile commands. It adds a new flag '--resetcache' which allows users to reset the cache and retrieve the latest profile permissions from the org. This fix addresses issue #1413.

The second patch fixes a formatting issue with the prettier in the sfpowerkit utility file.

Overall, these patches enhance the profile merge and reconcile commands and improve the formatting of the sfpowerkit utility.

#### Checklist

- [ ] Tested changes?

### DIFF
--- a/packages/sfpowerscripts-cli/src/commands/profile/retrieve.ts
+++ b/packages/sfpowerscripts-cli/src/commands/profile/retrieve.ts
@@ -65,7 +65,7 @@ export default class Retrieve extends SfpowerscriptsCommand {
             folders.push(...argFolder);
         }
 
-        Sfpowerkit.initCache();
+        Sfpowerkit.initCache(true);
 
         SFPLogger.log(COLOR_WARNING(messages.getMessage('retriveDelayWarning')),LoggerLevel.INFO);
         SFPLogger.log(COLOR_KEY_MESSAGE(`Retrieving profiles from ${this.flags.targetorg}`),LoggerLevel.INFO );
--- a/packages/sfprofiles/src/utils/sfpowerkit.ts
+++ b/packages/sfprofiles/src/utils/sfpowerkit.ts
@@ -30,8 +30,10 @@ export class Sfpowerkit {
             fs.unlinkSync(cachePath);
     }
 
-    public static initCache() {
+    public static initCache(resetCache?:boolean) {
         try {
+             if(resetCache)
+                Sfpowerkit.resetCache();
             //Set the cache path on init,
             //TODO: Move this to a temporary directory with randomization
             Sfpowerkit.cache = new SQLITEKeyValue(FileUtils.getLocalCachePath('sfpowerkit-cache.db'));
--- a/packages/sfprofiles/src/utils/sfpowerkit.ts
+++ b/packages/sfprofiles/src/utils/sfpowerkit.ts
@@ -3,9 +3,8 @@ import chalk from 'chalk';
 import * as fs from 'fs-extra';
 import SQLITEKeyValue from './sqlitekv';
 import FileUtils from './fileutils';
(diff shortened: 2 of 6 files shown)


packages/sfprofiles/src/utils/sfpowerkit.ts
```diff
@@ -30,8 +30,10 @@ export class Sfpowerkit {
             fs.unlinkSync(cachePath);
     }
 
-    public static initCache() {
+    public static initCache(resetCache?:boolean) {
         try {
+             if(resetCache)
```
Schuchie: @azlam-abdulsalam 
Incorrect line indentation

packages/sfpowerscripts-cli/src/commands/profile/retrieve.ts
```diff
@@ -65,7 +65,7 @@ export default class Retrieve extends SfpowerscriptsCommand {
             folders.push(...argFolder);
         }
 
-        Sfpowerkit.initCache();
+        Sfpowerkit.initCache(true);
```
Schuchie, line 68: @azlam-abdulsalam 
Why did you only optionally reseted the cache for the other two commands and forced it for the retrieve? Why not uniformly? The error can occur with every command. And is there anything against not flushing the cache when changing the org? Or to build the cache key with alias?

zhebinliu: same question here

azlam-abdulsalam: Reconcile is used in sfpowerscripts using a dual pass approach, so if we reset the second one will be expensive. Hence made optional, it makes sense though while you retrieve to ensure you ignore any existing cache.

 Ideally the fix would be to be have a proper data structure, that caches for every org, as you proposed, but I do think this is a  cheaper option for now to get this bug resolved, and then proceed to fix it later

Schuchie: For me thats fine fore a temp fix. How do you decide if a cache is valid or not anymore?

